PR #2, opened 2020-12-18
Remove log level

### LINKED ISSUE #1: log level in cli
After the introduction of the `--log_level` cli parameter, we lost the ability to filter logs by package making it unusable for log levels like debug or more granular (because there is too much output from package like rustls and sled).

example
```
RUST_LOG=debug,sled=info,rustls=info cargo run ...
```

I think we should either:
* remove `--log_level` parameter
* if possible accept in `--log_level` the same format supported in RUST_LOG and even allow RUST_LOG to work in alternative (I missed the `--log_level`update and I started trying to use RUST_LOG automatically which is the standard in the rust ecosystem and I was surprised it had no effect)

notmandatory:
### Description

This issue fixes #1 by removing the `--log_level` option which was overriding the `RUST_LOG` environment variable but does not provide the same level of logging control.

### Notes to the reviewers

Also in this PR I added a couple more usage examples in the README, including an example for using the `RUST_LOG` env variable.

#### All Submissions:

* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)

#### Bugfixes:

* [x] This pull request breaks the existing API

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -29,7 +29,7 @@ use std::sync::Arc;
 
 use bitcoin::Network;
 use clap::AppSettings;
-use log::{debug, info, warn, LevelFilter};
+use log::{debug, info, warn};
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
 use structopt::StructOpt;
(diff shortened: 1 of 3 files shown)


notmandatory:
@RCasatta I move #1 to the new `bdk-cli` repo and this is the fix for it.

src/repl.rs
```diff
@@ -71,9 +71,6 @@ fn prepare_home_dir() -> PathBuf {
 fn main() {
     let cli_opt: WalletOpt = WalletOpt::from_args();
 
-    let level = LevelFilter::from_str(cli_opt.log_level.as_str()).unwrap_or(LevelFilter::Info);
-    env_logger::builder().filter_level(level).init();
```
RCasatta, line 75: I think we need default initialization of the logger?

`env_logger::init();`

notmandatory: Oh ya! I should have caught that. Fixed now.

PR #5, opened 2020-12-28
Add network parser function to lib.rs, remove network parsing from repl.rs

notmandatory:
### Description

This PR moves network parsing to lib.rs and changes repl to return an error instead of defaulting to testnet if an invalid network name is given.

#### All Submissions:

#### New Features:

#### Bugfixes:

* [x] This pull request breaks the existing API

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@
 //!                     "sync", "--max_addresses", "50"];
 //! let cli_opt = WalletOpt::from_iter(&cli_args);
 //!
-//! let network = Network::from_str(cli_opt.network.as_str()).unwrap_or(Network::Testnet);
+//! let network = cli_opt.network;
 //!
 //! let descriptor = cli_opt.descriptor.as_str();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -400,6 +402,10 @@ fn parse_outpoint(s: &str) -> Result<OutPoint, String> {
     OutPoint::from_str(s).map_err(|e| format!("{:?}", e))
 }
 
+fn parse_network(s: &str) -> Result<Network, String> {
```
RCasatta, line 405: small doubt, can we avoid this since Network already has `from_str` -> Result<Network, Error>?

notmandatory: Good idea, I thought it needed a function that returned Result<Network,String> but just using `Network::from_str` works fine.

notmandatory:
BTW, I Removed commits to set `bdk` dependency version to ^0.2 as I don't think we're quite ready for that.

PR #8, opened 2021-01-05
Prep for publishing to crates.io, rename bin to bdkwallet

notmandatory:
### Description

In this PR I add some missing metadata to Cargo.toml to prepare for publishing to crates.io. I also renamed the `repl.rs` bin to `bdk-cli` to be consistent with similar tools in other projects.  

### Notes to the reviewers

See related PR to update the bitcoindevkit.org website docs to match updated `bdk-cli`.

#### All Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,14 @@
 name = "bdk-cli"
 version = "0.1.0"
 edition = "2018"
-authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>"]
+authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>", "Steve Myers <steve@notmandatory.org"]
+homepage = "https://bitcoindevkit.org"
+repository = "https://github.com/bitcoindevkit/bdk-cli"
+documentation = "https://docs.rs/bdk-cli"
+description = "A CLI library and example CLI tool based on the BDK descriptor-based wallet library"
+keywords = ["bitcoin", "wallet", "descriptor", "psbt"]
+readme = "README.md"
+license-file = "LICENSE"
 
 [dependencies]
 bdk = { git = "https://github.com/bitcoindevkit/bdk.git" }
@@ -25,8 +32,7 @@ repl = ["async-trait", "rustyline", "dirs-next", "env_logger", "clap", "electrum
 electrum = ["bdk/electrum"]
 esplora = ["bdk/esplora"]
 
-
 [[bin]]
-name = "repl"
-path = "src/repl.rs"
+name = "bdk-cli"
+path = "src/bdk_cli.rs"
 required-features = ["repl"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ readme = "README.md"
 license-file = "LICENSE"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


RCasatta:
I am not strongly against renaming, however, I prefer `bdk-cli` like other CLI tools such as `bitcoin-cli` or `lightning-cli`

btw, I think you have a typo in the description `bdwallet` instead of `bdkwallet`

notmandatory:
> I am not strongly against renaming, however, I prefer `bdk-cli` like other CLI tools such as `bitcoin-cli` or `lightning-cli`
> 
> btw, I think you have a typo in the description `bdwallet` instead of `bdkwallet`

Thanks for the feedback, consistency with similar tools makes sense.  I updated this PR to use `bdk-cli` and put in a corresponding PR to update the website to match.

notmandatory:
I also updated the app name in the `--help` to "BDK CLI".. last change I promise.

Cargo.toml
```diff
@@ -3,6 +3,13 @@ name = "bdk-cli"
 version = "0.1.0"
 edition = "2018"
 authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>"]
```
RCasatta, line 5: I think you should add yourself here

Cargo.toml
```diff
@@ -25,8 +32,7 @@ repl = ["async-trait", "rustyline", "dirs-next", "env_logger", "clap", "electrum
 electrum = ["bdk/electrum"]
 esplora = ["bdk/esplora"]
 
-
 [[bin]]
-name = "repl"
-path = "src/repl.rs"
+name = "bdk-cli"
+path = "src/bdkcli.rs"
```
RCasatta, line 37: nit: I would prefer bdk_cli.rs as the file name (while executable it's right with the hyphen)

notmandatory: Sounds good, fixed.

Cargo.toml
notmandatory: :pray: thanks guys, you did all the hard work but I'm happy to add my name and help keep the updates coming

notmandatory:
I added two more commits, one to set the bdk and bdk_macros versions so that I can push this crate to crates.io. And one to fix a broken rustdocs link.  If no objections I'll merge this PR tomorrow and then publish it.

Cargo.toml
```diff
@@ -2,11 +2,18 @@
...
 edition = "2018"
-authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>"]
+authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>", "Steve Myers <steve@notmandatory.org"]
+homepage = "https://bitcoindevkit.org"
+repository = "https://github.com/bitcoindevkit/bdk-cli"
+documentation = "https://docs.rs/bdk-cli"
+description = "A CLI library and example CLI tool based on the BDK descriptor-based wallet library"
+keywords = ["bitcoin", "wallet", "descriptor", "psbt"]
+readme = "README.md"
+license-file = "LICENSE"
```
RCasatta, line 12: bdk crate doesn't recognize the license this way: "non-standard"
We may want to express this way:

```license = "MIT"```

notmandatory: Oh ya, that is weird, I will fix it here and make a PR to fix it in `bdk` too.

PR #14, opened 2021-02-03
Re-organize wallet sub-commands and add `key` sub-commands

### LINKED ISSUE #12: Add subcommand to break out of the repl
Just for completeness. After a few minutes in the repl I tried exiting and ended up realizing that there were no commands for it and I had to ctrl+d out.

An `exit` subcommand would be a nice touch.

notmandatory:
### Description

Depends on https://github.com/bitcoindevkit/bdk/pull/274

Added in Lib
- `CliOpts` struct and `CliSubCommand` enum representing top level cli options and commands
- `KeySubCommand` enum
- `handle_key_subcommand` function

Changed in Lib
- Renamed `WalletOpt` struct to `WalletOpts`
- `WalletSubCommand` enum split into `OfflineWalletSubCommand` and `OnlineWalletSubCommand`
- Split `handle_wallet_subcommand` into two functions, `handle_offline_wallet_subcommand` and `handle_online_wallet_subcommand`
- A wallet without a `Blockchain` is used when handling offline wallet sub-commands

Added in `bdk-cli`
- Top level commands "wallet", "key", and "repl"
- "key" command has sub-commands to "generate" and "restore" a master extended key
- "repl" command now has an "exit" sub-command, fixes #12 

Changed in `bdk-cli`
- "wallet" sub-commands and options must be proceeded by "wallet" command
- "repl" command loop now includes both "wallet" and "key" sub-commands

### Notes to the reviewers

There's a little bug in the help / about info while in REPL mode. ~~I don't think there's a way to work around it but there is a [`structopt` issue] for it.~~ I found a [workaround] for this issue.

[`structopt` issue]: https://github.com/TeXitoi/structopt/issues/391
[workaround]: https://github.com/TeXitoi/structopt/issues/333#issuecomment-712265332

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { git = "https://github.com/afilini/bdk.git", branch = "fix/better-derivable-key-api", features = ["all-keys"]}
-bdk-macros = "^0.2"
+bdk = { git = "https://github.com/bitcoindevkit/bdk.git", commit = "4c36020e", features = ["all-keys"]}
+bdk-macros = { git = "https://github.com/bitcoindevkit/bdk.git", commit = "4c36020e" }
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
 log = "^0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { version = "^0.3", default-features = false }
+bdk = { git = "https://github.com/afilini/bdk.git", branch = "fix/better-derivable-key-api", features = ["all-keys"]}
 bdk-macros = "^0.2"
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
@@ -25,10 +25,11 @@ rustyline = { version = "6.0", optional = true }
 dirs-next = { version = "2.0", optional = true }
 env_logger = { version = "0.7", optional = true }
 clap = { version = "2.33", optional = true }
+regex = {version = "1", optional = true }
 
 [features]
-default = ["repl", "esplora", "electrum", "bdk/key-value-db"]
-repl = ["async-trait", "rustyline", "dirs-next", "env_logger", "clap", "electrum"]
+default = ["repl", "electrum"]
+repl = ["async-trait", "rustyline", "dirs-next", "env_logger", "clap", "regex"]
 electrum = ["bdk/electrum"]
 esplora = ["bdk/esplora"]
 compiler = ["bdk/compiler"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { git = "https://github.com/bitcoindevkit/bdk.git", commit = "4c36020e", features = ["all-keys"]}
-bdk-macros = { git = "https://github.com/bitcoindevkit/bdk.git", commit = "4c36020e" }
+bdk = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179", features = ["all-keys"]}
+bdk-macros = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179" }
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
 log = "^0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179", features = ["all-keys"]}
+bdk = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179", default-features = false, features = ["all-keys"]}
 bdk-macros = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179" }
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
@@ -28,8 +28,8 @@ clap = { version = "2.33", optional = true }
 regex = {version = "1", optional = true }
 
 [features]
-default = ["repl", "electrum"]
-repl = ["async-trait", "rustyline", "dirs-next", "env_logger", "clap", "regex"]
+default = []
+repl = ["async-trait", "bdk/key-value-db", "clap", "dirs-next", "env_logger", "regex", "rustyline"]
 electrum = ["bdk/electrum"]
 esplora = ["bdk/esplora"]
 compiler = ["bdk/compiler"]
@@ -37,4 +37,7 @@ compiler = ["bdk/compiler"]
 [[bin]]
 name = "bdk-cli"
 path = "src/bdk_cli.rs"
-required-features = ["repl"]
+required-features = ["repl", "electrum"]
+
+[package.metadata.docs.rs]
+all-features = true
--- a/src/bdk_cli.rs
+++ b/src/bdk_cli.rs
@@ -50,9 +50,11 @@ use bdk_cli::{
 };
 use regex::Regex;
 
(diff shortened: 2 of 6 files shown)


thunderbiscuit:
ACK.

I have tested this locally and everything works well! I have not reviewed the code, however.

notmandatory:
I added a small change to `bdk_cli.rs` to remove the unneeded `ReplOpt` struct.

afilini:
Reviewing it now :)

src/bdk_cli.rs
```diff
@@ -100,34 +109,95 @@ fn main() {
...
 
-    match cli_opt.subcommand {
-        WalletSubCommand::Repl => {
             let mut rl = Editor::<()>::new();
 
             // if rl.load_history("history.txt").is_err() {
             //     println!("No previous history.");
             // }
 
+            let split_regex = Regex::new(r#"[\w\-]+|"[\w\s]*""#).unwrap();
```
afilini, line 198: If I understand this correctly, the goal here is to split at every space except when the space is contained in the double quotes, right?

afilini: And what does the `filter_regex` do then?

afilini: Oh, is it to remove the quotes?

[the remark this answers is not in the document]
afilini: We could also add support for single quotes, should be trivial at this point

src/lib.rs
```diff
@@ -402,30 +544,23 @@ fn parse_outpoint(s: &str) -> Result<OutPoint, String> {
...
+/// Execute an offline wallet sub-command
 ///
-/// Wallet sub-commands are described in [`WalletSubCommand`]. See [`crate`] for example usage.
+/// Offline wallet sub-commands are described in [`OfflineWalletSubCommand`].
 #[maybe_async]
-pub fn handle_wallet_subcommand<C, D>(
-    wallet: &Wallet<C, D>,
-    wallet_subcommand: WalletSubCommand,
+pub fn handle_offline_wallet_subcommand<D>(
+    wallet: &Wallet<(), D>,
```
afilini, line 552: You could change this to take a generic type `T` instead of just `()`, which would allow you to call this with a ref to an online wallet

```suggestion
pub fn handle_offline_wallet_subcommand<T, D>(
    wallet: &Wallet<T, D>,
```

src/bdk_cli.rs
```diff
@@ -100,34 +109,95 @@ fn main() {
...
+            let result = bdk_cli::handle_key_subcommand(network, key_subcommand);
+            serde_json::to_string_pretty(&result.unwrap()).unwrap()
+        }
+        CliSubCommand::Repl { wallet_opts } => {
+            let database = open_database(&wallet_opts);
+            let online_wallet = new_online_wallet(network, &wallet_opts, database.clone()).unwrap();
+            let online_wallet = Arc::new(online_wallet);
+
+            let offline_wallet = new_offline_wallet(network, &wallet_opts, database).unwrap();
+            let offline_wallet = Arc::new(offline_wallet);
```
afilini, line 190: ... then you could remove this and just use the `online_wallet` everywhere

src/bdk_cli.rs
```diff
@@ -136,22 +206,46 @@ fn main() {
...
-                            repl_subcommand.unwrap().subcommand,
-                        )
-                        .unwrap();
-                        println!("{}", serde_json::to_string_pretty(&result).unwrap());
+                        let repl_subcommand = repl_subcommand.unwrap();
+
+                        let result = match repl_subcommand {
+                            ReplSubCommand::OnlineWalletSubCommand(online_subcommand) => {
+                                bdk_cli::handle_online_wallet_subcommand(
+                                    &Arc::clone(&online_wallet),
```
afilini, line 229: I think you could just use `&online_wallet` here and in all the other places

Cargo.toml
```diff
@@ -12,8 +12,8 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { version = "^0.3", default-features = false }
-bdk-macros = "^0.2"
+bdk = { git = "https://github.com/bitcoindevkit/bdk.git", rev = "c4f2179", features = ["all-keys"]}
```
afilini, line 15: Needs `default-features = false`, otherwise it'll include `electrum` by default which can't be compiled under `wasm32`

src/bdk_cli.rs
afilini: Actually, I think you could remove the `Arc` stuff entirely.. I don't think we are using them anymore, we are just passing normal references around

src/lib.rs
```diff
@@ -582,86 +703,302 @@ where
...
+    let secp = Secp256k1::new();
+
+    match subcommand {
+        KeySubCommand::Generate {
+            word_count,
+            password,
+        } => {
+            let mnemonic_type = match word_count {
+                12 => MnemonicType::Words12,
+                _ => MnemonicType::Words24,
```
afilini, line 805: Maybe return an error if an invalid number is passed? I think it's fine to only support 12 and 24, but if I pass like 7 I'd like to see an error rather than it silently use 24.

afilini (changes requested):

src/bdk_cli.rs
notmandatory: Hi yes the filter regex removes the surrounding quotes

src/bdk_cli.rs
```diff
@@ -136,22 +206,46 @@ fn main() {
                             continue;
                         }
                         rl.add_history_entry(line.as_str());
-                        let split_line: Vec<&str> = line.split(' ').collect();
-                        let repl_subcommand: Result<ReplOpt, clap::Error> =
-                            ReplOpt::from_iter_safe(split_line);
+                        let split_line: Vec<&str> =
+                            split_regex.find_iter(&line).map(|m| m.as_str()).collect();
```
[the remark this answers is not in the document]
notmandatory: oh cool, I'll give it a try!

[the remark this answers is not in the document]
notmandatory: yes this is much better, I also added the capture group you suggested for single quotes, and fixed it so the double quote and single quote capture groups can take non-alphanumeric characters (other than " or ')

src/lib.rs
```diff
@@ -104,60 +105,75 @@ use std::str::FromStr;
...
+///                     #[cfg(feature = "electrum")]
+///                     retries: 5,
+///                     #[cfg(feature = "electrum")]
+///                     timeout: None,
+///                     #[cfg(feature = "electrum")]
+///                     electrum: "ssl://electrum.blockstream.info:60002".to_string(),
+///                     #[cfg(feature = "esplora")]
+///                     esplora: None,
+///                     #[cfg(feature = "esplora")]
+///                     esplora_concurrency: 4,
```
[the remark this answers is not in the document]
notmandatory: Yes good call, much cleaner this way. I originally had these two structs but took them out when I was trouble shooting some `structopt` help issues. Back in now.

Cargo.toml
notmandatory: Fixed. Had to tweek docs tests a bit but should all work now.

src/bdk_cli.rs
notmandatory: I removed the Arc and it works fine.

src/bdk_cli.rs
notmandatory: I added the generic type to allow  `handle_offline_wallet_subcommand` to accept an online wallet and works great, removed the `offline_wallet` for REPL mode. 

I'm going to keep the offline_wallet for the offline commands when not in REPL mode. My goal is to eventually make `offline`, `online`, and `repl` optional cargo features. So someone could make a `bdk-cli` build to only do offline CLI commands. Or maybe only offline via CLI or REPL, etc.

src/lib.rs
notmandatory: Actually this is taken care of by `structopt`, if I pass an invalid value like 7 I get this message:
```
cargo run --all-features key generate -e 7
    ...
error: '7' isn't a valid value for '--entropy <WORD_COUNT>'
        [possible values: 12, 24]

```

afilini (approved):

src/bdk_cli.rs
afilini: Yes, sounds good!

PR #16, opened 2021-02-17
Various dependency fixups

LLFourn:
This fixes various dependency/feature things that were bothering me when porting old code to using this. See each commit for explanation :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ log = "^0.4"
 base64 = "^0.11"
 
 # Optional dependencies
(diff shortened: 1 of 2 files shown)


notmandatory:
Looks good, but I should have merged before changing the CI pipeline to use `1.50.0` instead of `stable`. Can you do a quick re-basing so it can pass the new required checks? Thanks!

notmandatory:
@LLFourn if you don't mind I'm happy to rebase this for you, I can do it as part of the merge.

PR #17, opened 2021-02-21
Add key sub-command to 'derive' xpub

notmandatory:
### Description

This PR updates dependencies `bdk` to `0.4.0` and `bdk-macros` to `0.3.0` and adds a new `key` `derive` subcommand. The `key` `derive` sub-command derives an xpub from a given master xprv and derivation path.  

The purpose of the `derive` command is to make it easier to create xpubs for use in testing multisig descriptors. For example an xpub can be derived and stored in an environment variable like this:

```shell
echo \"$ALICE_XPRV\"
"tprv8ZgxMBicQKsPfQjJy8ge2cvBfDjLxJSkvNLVQiw7BQ5gTjKadG2rrcQB5zjcdaaUTz5EDNJaS77q4DzjqjogQBfMsaXFFNP3UqoBnwt2kyT"
export ALICE_XPUB=$(bdk-cli key derive --xprv $ALICE_XPRV --path "m/84'/1'/0'/0" | jq -r ".xpub" | sed 's/^\[.*\]//g')
echo \"$ALICE_XPUB\"
"tpubDDrcq8JNTLVwaGnXp7KqZZxFJoic49ikEotWnpm6hpMY2hL2F1hpgbBBkEdsFJHU1iAx9DzMUKYr5mrphBMoXxhsBPhnVdmaoSCaeh6QWgj/0/*"
```

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `handle_key_subcommand` function
 
 #### Changed
-- Upgrade BDK to `0.3`
+- Upgrade `bdk` to `0.4.0` and `bdk-macros` to `0.3.0`
 - Renamed `WalletOpt` struct to `WalletOpts`
 - `WalletSubCommand` enum split into `OfflineWalletSubCommand` and `OnlineWalletSubCommand`
 - Split `handle_wallet_subcommand` into two functions, `handle_offline_wallet_subcommand` and `handle_online_wallet_subcommand`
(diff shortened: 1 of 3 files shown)


thunderbiscuit:
A few questions on this. Does the `derive` subcommand always create an `xpub` at the last node in the provided `--path`?

If that is the case (say you give the path to be `--path "m/84h/1h/0h/`, would it then return an `xpub` that then needs to be extended (to add the `account 0` and the `*`? In this case it ends up being 2 steps no matter what no?

notmandatory:
Yes, `derive` right now only creates an extended pub key and it always creates to the last hardened part of the path.. so if you use `--path "m/84'/1'/0'/123"` you'll get an `tpub..../123/* `, but if you use `--path "m/84'/1'/0'"` then the result is `tpub..../*`.  So it's just deriving the minimum that's needed (since hardened parts of the path can only be derived from the extended private key.

thunderbiscuit:
Ok I see! `xpub` happens at the last hardened node of the path, plus whatever extra derivation you want added on top. Thanks!

thunderbiscuit:
I'm wondering if removing the default BIP32 root node `xpub` returned by the `key generate` subcommand should be part of this PR as well. The two feel related since this `derive` is a more flexible and better option for xpubs, and keeping the default one there in the `generate` subcommand only leads to more confusion IMO.

notmandatory:
oh ya I should have mention it but I also removed the root xpub from the output of the `generate` and `restore` commands. :-)

PR #22, opened 2021-03-10
Make repl and electrum default features

RCasatta:
### Description

Make repl and electrum default features and updates readme accordingly (also add a specific example for esplora)

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #23, opened 2021-03-11
Remove unwraps in bdk_cli and more detailed message for ChecksumMismatch

RCasatta:
### Description

This removes some unwraps in favor of ? during handling of the commands.

During development I frequently hit ChecksumMismatch while changing the descriptor and keeping the same wallet name, I imagined users may be confused so I detailed the message.

### Notes to the reviewers

On top of #22 

Error handling is a little nicer for end-user but it comes at the expense of losing stacktrace information in case of errors, so it may be controversial. However, we are now writing blog post and expanding user base so I think it is better.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 - `CliSubCommand::Compile` enum variant and `handle_compile_subcommand()` function
 - Make repl and electrum default features
+- Remove unwraps while handling CLI commands
 
 ### `bdk-cli` bin
 
--- a/src/bdk_cli.rs
+++ b/src/bdk_cli.rs
@@ -27,7 +27,7 @@ use std::path::PathBuf;
 
 use bitcoin::Network;
 use clap::AppSettings;
-use log::{debug, info, warn};
+use log::{debug, error, info, warn};
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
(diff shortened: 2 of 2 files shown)


thunderbiscuit:
Great idea! The `ChecksumMismatch` error message has been bugging me for some time now, and I think a more user-friendly explanation as to what is going on is a really good thing.

Cargo.toml
```diff
@@ -27,7 +27,7 @@ clap = { version = "2.33", optional = true }
 regex = {version = "1", optional = true }
 
 [features]
-default = []
+default = ["repl", "electrum"]
```
notmandatory, line 30: The reason we didn't add any default features is because the `wasm` playground app doesn't need `repl` and can't use `electrum` .. but maybe this is OK if we just build `bitcoindevkit.org/playground` with `default-features = false` and `features=["esplora"]`.  @afilini does that sound ok?

notmandatory:
At some point I'd like to try adding [anyhow](https://crates.io/crates/anyhow) to this project to add context to and help display errors, but I think we'd also need to add [thiserror](https://crates.io/crates/thiserror) macro for `bdk` errors. For now I think the solution in the PR is a good step in the right direction.

Cargo.toml
afilini: Yes, exactly. As long as there's a way to disable them it should be fine

afilini:
This one probably needs rebasing after the force-push on #22

RCasatta:
rebased

PR #26, opened 2021-05-08
Compact filter

rajarshimaitra:
### Description

This PR adds the compact filter configuration to connect bdk-cli with a Bitcoin Core node serving compact filters.

### Note to Reviewers
I have updated the Readme doc with new usage instruction. If thats not clear (because github doesn't show .md outputs in PR) below is the instruction:

To sync a wallet to Bitcoin Core node (assuming a regtest node at 127.0.0.1:18444) serving compact filters:

Then run:
```shell
cargo run --features compact_filters -- --network regtest wallet --node 127.0.0.1:18444 --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
```
   

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 - New top level command "Compile" which compiles a miniscript policy to an output descriptor
+- `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration 
 
 ## [0.2.0]
 
@@ -42,7 +43,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - "key" sub-commands to "generate" and "restore" a master private key
 - "key" sub-command to "derive" an extended public key from a master private key
 - "repl" command now has an "exit" sub-command
-- `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration 
 
 #### Changed
 - "wallet" sub-commands and options must be proceeded by "wallet" command
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - "key" sub-commands to "generate" and "restore" a master private key
 - "key" sub-command to "derive" an extended public key from a master private key
 - "repl" command now has an "exit" sub-command
+- `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration 
 
 #### Changed
 - "wallet" sub-commands and options must be proceeded by "wallet" command
--- a/README.md
+++ b/README.md
@@ -23,6 +23,18 @@ To sync a wallet to the default electrum server:
 cargo run -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
 ```
 
+To sync a wallet to Bitcoin Core node (assuming a regtest node at 127.0.0.1:18444) serving compact filters:
+
+Update `Cargo.toml`
+```
+[features]
+default = ["repl", "electrum", "compact_filters"]
+```
+Then run:
+```shell
+cargo run -- --network regtest wallet --node 127.0.0.1:18444 --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
+```
+
 To get a wallet balance with customized logging:
 
 ```shell
--- a/README.md
+++ b/README.md
@@ -24,15 +24,12 @@ cargo run -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN
 ```
 
 To sync a wallet to Bitcoin Core node (assuming a regtest node at 127.0.0.1:18444) serving compact filters:
+Note: 
+- This will increase build time by few minutes for the binaries because of `librocksdb`.
+- Bitcoin Core v0.21.0 or higher is required to serve compact filters.  
 
-Update `Cargo.toml`
-```
-[features]
-default = ["repl", "electrum", "compact_filters"]
-```
-Then run:
 ```shell
-cargo run -- --network regtest wallet --node 127.0.0.1:18444 --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
+cargo run --features compact_filters -- --network regtest wallet --node 127.0.0.1:18444 --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
 ```
 
 To get a wallet balance with customized logging:
--- a/src/bdk_cli.rs
+++ b/src/bdk_cli.rs
@@ -32,6 +32,8 @@ use rustyline::error::ReadlineError;
 use rustyline::Editor;
 use structopt::StructOpt;
 
+#[cfg(feature = "compact_filters")]
+use bdk::blockchain::compact_filters::{BitcoinPeerConfig, CompactFiltersBlockchainConfig};
 #[cfg(feature = "esplora")]
 use bdk::blockchain::esplora::EsploraBlockchainConfig;
 use bdk::blockchain::{
@@ -117,8 +119,32 @@ where
         timeout: wallet_opts.electrum_opts.timeout,
     });
 
-    // Fall back to Electrum config if Esplora config isn't provided
-    let config = config_esplora.unwrap_or(config_electrum);
+    #[cfg(feature = "compact_filters")]
+    let compact_filter_config: Option<AnyBlockchainConfig> = {
+        let peerconfig = BitcoinPeerConfig {
+            address: wallet_opts.compactfilter_opts.address.clone(),
+            socks5: None,
+            socks5_credentials: None,
+        };
+
+        Some(AnyBlockchainConfig::CompactFilters(
+            CompactFiltersBlockchainConfig {
+                peers: vec![peerconfig],
+                network,
+                storage_dir: wallet_opts.wallet.clone(),
+                skip_blocks: wallet_opts.compactfilter_opts.skip_blocks,
+            },
+        ))
+    };
+
+    #[cfg(not(feature = "compact_filters"))]
+    let compact_filter_config = None;
+
+    // Fall back to Electrum config if Esplora or Compact Filter config isn't provided
+    let config = config_esplora.unwrap_or_else(|| match compact_filter_config {
+        Some(config) => config,
+        None => config_electrum,
+    });
 
     let descriptor = wallet_opts.descriptor.as_str();
     let change_descriptor = wallet_opts.change_descriptor.as_deref();
--- a/src/bdk_cli.rs
+++ b/src/bdk_cli.rs
@@ -79,7 +79,11 @@ fn prepare_home_dir() -> PathBuf {
         fs::create_dir(&dir).unwrap();
     }
 
+    #[cfg(not(feature = "compact_filters"))]
     dir.push("database.sled");
+
+    #[cfg(feature = "compact_filters")]
+    dir.push("compact_filters");
     dir
 }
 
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - "key" sub-commands to "generate" and "restore" a master private key
 - "key" sub-command to "derive" an extended public key from a master private key
 - "repl" command now has an "exit" sub-command
+- `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration
```
afilini, line 45: This should be moved up under the "unreleased" section, since it will become part of the following release

README.md
```diff
@@ -23,6 +23,18 @@ To sync a wallet to the default electrum server:
 cargo run -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
 ```
 
+To sync a wallet to Bitcoin Core node (assuming a regtest node at 127.0.0.1:18444) serving compact filters:
```
afilini, line 26: Since it takes a very long time to compile `librocksdb` I would add a little warning here in the README. Something like "keep in mind that enabling this feature will make building the projects take a few more minutes, as compact filters require RocksDB which takes a while to build".

src/bdk_cli.rs
```diff
@@ -117,8 +119,32 @@ where
...
+    let compact_filter_config: Option<AnyBlockchainConfig> = {
+        let peerconfig = BitcoinPeerConfig {
+            address: wallet_opts.compactfilter_opts.address.clone(),
+            socks5: None,
+            socks5_credentials: None,
+        };
+
+        Some(AnyBlockchainConfig::CompactFilters(
+            CompactFiltersBlockchainConfig {
+                peers: vec![peerconfig],
```
afilini, line 132: To make it sync faster you can connect multiple times to the same peer. I would add an extra option to `CompactFilterOpts` with the number of "cores" to use in parallel to sync. We could default to 2 or 4 which is reasonable on modern hardware

src/bdk_cli.rs
```diff
@@ -117,8 +119,32 @@ where
...
+                network,
+                storage_dir: wallet_opts.wallet.clone(),
```
afilini, line 134: Ideally this should end up in the same data directory that we use for the database. Maybe you can refactor `prepare_home_dir()` a little bit to reuse the same path (currently it always appends `database.sled` at the end but obviously you don't want that for the compact filters dir)

notmandatory (changes requested):

src/bdk_cli.rs
```diff
@@ -117,8 +119,32 @@ where
         timeout: wallet_opts.electrum_opts.timeout,
     });
 
-    // Fall back to Electrum config if Esplora config isn't provided
-    let config = config_esplora.unwrap_or(config_electrum);
+    #[cfg(feature = "compact_filters")]
+    let compact_filter_config: Option<AnyBlockchainConfig> = {
```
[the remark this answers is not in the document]
rajarshimaitra: Done.

src/bdk_cli.rs
```diff
@@ -117,8 +119,32 @@ where
...
+        let peerconfig = BitcoinPeerConfig {
+            address: wallet_opts.compactfilter_opts.address.clone(),
+            socks5: None,
```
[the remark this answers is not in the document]
rajarshimaitra: Noted. Also adding SOCKS5 optional credentials. 

Because of so many options, flag reuse clashes are happening. So had to remove `short` for some cases. Or we can use some other alphabets, but might not make intuitive sense.

src/bdk_cli.rs
```diff
@@ -117,8 +119,32 @@ where
...
+    };
+
+    #[cfg(not(feature = "compact_filters"))]
+    let compact_filter_config = None;
+
+    // Fall back to Electrum config if Esplora or Compact Filter config isn't provided
+    let config = config_esplora.unwrap_or_else(|| match compact_filter_config {
+        Some(config) => config,
+        None => config_electrum,
+    });
```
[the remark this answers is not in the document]
rajarshimaitra: Thanks, ya much better.

CHANGELOG.md
rajarshimaitra: eh. I should have seen that. Updated in revised.

README.md
rajarshimaitra: It did take longer in my machine, but nor considerably. Adding the note. @notmandatory noted. Updated in revised.

src/bdk_cli.rs
rajarshimaitra: Yes, that would make sense, but `BitcoinPeerConfig` isn't `Clone` so we cant just copy same config n number of times.

I was trying to see if instead of giving one address, we can provide multiple addresses in cli which then will be converted into `Vec<Peer>`. But doesn't seem to be a easy way to do that with `structopt`. 

If we make the `BitcoinPeerConfig` `Clone` in upstream. then we can add multiuple connection to a single peer here.

Maybe I can do that in a separate PR.

src/bdk_cli.rs
rajarshimaitra: Not sure if I get it. Are you suggesting we should have different data directory for `electrum` `esplora` and `compact_filters`?

That would make sense. Currently it uses `database.sled` directory for all the cases in `.bdk-bitcoin` dir. We can have separate directories for each cases.

README.md
```diff
@@ -23,6 +23,18 @@ To sync a wallet to the default electrum server:
...
+
+Update `Cargo.toml`
```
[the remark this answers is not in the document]
rajarshimaitra: Yes that makes more sense. Updated.

src/lib.rs
```diff
@@ -324,6 +324,24 @@ pub struct WalletOpts {
...
+#[cfg(feature = "compact_filters")]
+#[derive(Debug, StructOpt, Clone, PartialEq)]
+pub struct CompactFilterOpts {
+    /// Sets the light client network address
+    #[structopt(name = "ADDRESS:PORT", short = "n", long = "node")]
+    pub address: String,
+
+    /// Optionally skip initial `skip_blocks` blocks (default: 0)
+    #[structopt(name = "SKIP_BLOCKS", short = "k", long = "skip_blocks")]
+    pub skip_blocks: Option<usize>,
```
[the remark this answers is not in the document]
rajarshimaitra: Updated.

rajarshimaitra:
Updated with the requested changes.

I am hitting a `InvalidHeaders` error while trying to connect bdk-cli to regtest node.

Not sure whats going wrong. Happening also in previous revision and BDK upstream.

rajarshimaitra:
Updated the `database_path` to include `wallet-name` provided with cli args, or use the default name. 

The objective is to keep multiple wallet data directories separate.

This PR is ready for a second look.

src/lib.rs
```diff
@@ -324,6 +324,43 @@ pub struct WalletOpts {
...
+    pub compactfilter_opts: CompactFilterOpts,
+}
+
+/// Compact Filter options
+///
+/// Compact filter peer information used by [`OnlineWalletSubCommand`]s.
+#[cfg(feature = "compact_filters")]
+#[derive(Debug, StructOpt, Clone, PartialEq)]
+pub struct CompactFilterOpts {
+    /// Sets the SOCKS5 proxy for the Electrum client
```
notmandatory, line 338: The doc comment here should be something like:
Sets the SOCKS5 proxy for the compact-filter blockchain client

src/lib.rs
```diff
@@ -324,6 +324,43 @@ pub struct WalletOpts {
...
+    #[structopt(name = "CPROXY_SERVER:PORT", long = "cmpct_proxy")]
+    pub proxy: Option<String>,
+
+    #[structopt(name = "PROXY_SERVER:USER", short = "u", long = "user")]
```
notmandatory, line 342: Need to add docs so they show up in the --help output, something like:
If --cmpct_proxy is set, sets the SOCKS5 proxy user name

src/lib.rs
```diff
@@ -324,6 +324,43 @@ pub struct WalletOpts {
...
+    pub user: Option<String>,
+
+    #[structopt(name = "PROXY_SERVER:PASSWD", long = "passwd")]
+    pub passwd: Option<String>,
+
+    /// Sets the light client network address
+    #[structopt(name = "ADDRESS:PORT", short = "n", long = "node")]
+    pub address: String,
+
+    /// Sets the number of parallel node connection
```
notmandatory, line 352: nit: "connections", should be plural

notmandatory, line 352: nit: "connections", should be plural

src/lib.rs
```diff
@@ -324,6 +324,43 @@ pub struct WalletOpts {
...
+}
+
+/// Compact Filter options
+///
+/// Compact filter peer information used by [`OnlineWalletSubCommand`]s.
+#[cfg(feature = "compact_filters")]
+#[derive(Debug, StructOpt, Clone, PartialEq)]
+pub struct CompactFilterOpts {
+    /// Sets the SOCKS5 proxy for the Electrum client
+    #[structopt(name = "CPROXY_SERVER:PORT", long = "cmpct_proxy")]
```
notmandatory, line 339: Since both electrum and cbf take an optional proxy I think it'd be clearer to move the proxy options from ElectrumOpts (proxy, retries, timeout) ,and from CompactFilterOpts (proxy, user, passwd) and combine into a new ProxyOpts struct. Then add a ProxyOpts field to WalletOpts and enable it when electrum or compact_filters features are enabled. The specific unused proxy opts can be ignored, possibly with a warning.. ie. the user supplies a  user/password for the electrum client.  Also to reduce the number of opts.. I think you could combine user and passwd into proxy_auth that takes a "PROXY_USER:PASSWD". Ideally in the future we'll be able to get all options supported for any blockchain client.

src/lib.rs
```diff
@@ -324,6 +324,43 @@ pub struct WalletOpts {
...
+    pub proxy: Option<String>,
+
+    #[structopt(name = "PROXY_SERVER:USER", short = "u", long = "user")]
+    pub user: Option<String>,
+
+    #[structopt(name = "PROXY_SERVER:PASSWD", long = "passwd")]
+    pub passwd: Option<String>,
+
+    /// Sets the light client network address
+    #[structopt(name = "ADDRESS:PORT", short = "n", long = "node")]
```
notmandatory, line 349: Similar to ElectrumOpts server, I'd set a `default_value` for CompactFiltersOpts address ("node") to "127.0.0.1:18444", hopefully in the future we'll be able to make this optional and find our peers via dns seed discovery if none are given. But for now I think it's ok to just default to what most people will probably be testing with. This also give an example to show users the expected format.  I'd also add to the description to make it clear this is peer full node address, something like: "Sets the compact block filter enabled full node address"

src/lib.rs
rajarshimaitra: Yes, thanks this makes sense. 

Going with a dedicated `ProxyOpts` with `PROXY_USER:PASSWD` pattern.

src/lib.rs
rajarshimaitra: Yes agreed. Also enabling `conn_coun` to start multiple connections to the same node. This parameter is not used right now. 

Also thinking about having `--node` as a list of `ADDRS:PORT` strings. Instead of one address, user can give multiple addresses and wallet will make a total of `no. of address * conn_coun` connections.

Does that make sense?

src/lib.rs
rajarshimaitra: Fixed.

src/lib.rs
rajarshimaitra: Fixed.

src/lib.rs
notmandatory: Sounds good but I don't think anyone has tested the CBF blockchain code with multiple nodes, just with multiple connections to the same node. But seems like it should work and maybe now is a good time to test it. Worst case we go back to just connecting to a single node multiple times.

rajarshimaitra: Ya. Updated the code. Planning to run that exact test too. Even if a multiple node connection doesn't work, the user can still give a single node address and it will resume standard behavior.

rajarshimaitra:
Updated with requested changes. Code is ready for further testing and review.

rajarshimaitra:
### Update for Reviewers

There is a likely error like this while trying to connect to a full node.
```
$ ./target/debug/bdk-cli --network regtest wallet --node 127.0.0.1:18444 --descriptor $desc sync
[2021-05-26T14:18:18Z ERROR bdk_cli] CompactFilters(InvalidHeaders)
```

This happens while connecting to fresh instant of `regtest-box` while having the old `~/.bdk-bitcoin/compact_filters` datadir. The previous filters are not valid for the new regtest node.

Solution is to to delete the datadir, then run.

src/bdk_cli.rs
rajarshimaitra: This is solved by creating multiple `BitcoinPeerConfig` for each connection instead of cloning.

src/bdk_cli.rs
rajarshimaitra: datadir updated as `~/.bdk-bitcoin/compact_filters/<wallet-name>`

src/bdk_cli.rs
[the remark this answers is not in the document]
rajarshimaitra: Proxy server configuration is extracted away into a separate `ProxyOpt` structure.

notmandatory:
Looks like the `electrum` test is failing, I can show you how to spin up docker to run the tests. The steps are also on the bitcoin-regtest-box README under [Local `bdk` testing](https://github.com/bitcoindevkit/bitcoin-regtest-box#local-bdk-testing).

rajarshimaitra:
@notmandatory Yes, the doc tests were failing as I forgot to change them to be compatible with the current `WalletOpts` structure. The tests should pass now.

notmandatory (approved):

notmandatory (approved):
ACK f22828d

PR #28, opened 2021-05-27
adapt to the changed signing interface in bdk 0.7

ulrichard:
### Description

Adapting to the current master branch of bdk.

After the next release of bdk, Cargo.toml, should go back to use crates.io

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -10,8 +10,8 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.50.0 # STABLE
-          - 1.45.0 # MSRV
+          - 1.51.0 # STABLE
+          - 1.46.0 # MSRV
         features:
           - default
           - repl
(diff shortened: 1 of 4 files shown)


notmandatory:
I'd prefer to just update to the latest published `bdk` version `^0.7` which should have all the latest and greatest features. Or `^0.8` will be out soon if there's something you need that's not released yet.

ulrichard:
Ah yes. I was not aware that the changes of the signing functions are in a release already.

ulrichard:
The CI failure is with rust 1.45.
In the BDK repo 1.46 is used instead. 
Wouldn't it make sense to use the same versions.
And if so, where can this be adapted?

notmandatory:
The reason `bdk` MSRV (minimum supported rust version) was changed to 1.46 is due to an issue with one of the dependencies (hyper), for more details see https://github.com/bitcoindevkit/bdk/pull/317.  

But to be in sync with `bdk` I don't see a problem incrementing the MSRV for `bdk-cli` also to 1.46. You just need to update it on line 14 of `cont_integration.yml`.

notmandatory:
Thanks for the updates. I  had to force-push to rebase on the lastest master changes. I also fixed up the `CHANGELOG.md`.

Issue #32, opened 2021-07-12
Add verbose (v) flag to the wallet subcommand, if enabled display PSBTs also as JSON 


notmandatory:
Currently the "wallet" subcommands "create_tx", "sign", and "finalize_psbt" returned a PSBT in serialized format. Add a new wallet flag with long name "verbose" and short name "v" that when enabled with cause these sub-sub-commands to also return the PSBT in deserialized JSON format, in addition to the existing serialized form currently returned.

This is a good first issue for @sandipndev

notmandatory:
Fixed via #33

PR #33, opened 2021-07-14
Wallet Verbosity for PSBT details

sandipndev:
Added "wallet" flag `-v` or `--verbose` for the following offline wallet subcommands: "create_tx", "sign" and "finalize_psbt".

Passed `walletOpts` down to `handle_offline_sub_commands` so that the subcommands can access and know if the verbose flag was set in wallet level.

This PR addresses the issue brought up in #32.


notmandatory (approved):
ACK. I also manually tested and works as expected.

PR #36, opened 2021-08-03
Draft RPC backend implementation

rajarshimaitra:
### Description
Draft work to expose RPC backend of BDK into bdk-cli. 

This is WIP and opened to have approach review and suggestions.  

#### All Submissions:

* [ ] I've signed all my commits
* [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature
* [ ] I've updated `CHANGELOG.md`

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #37, opened 2021-08-03
Add wallet option to select which blockchain client to use

notmandatory:
### Description

Added the `--blockchain_client` or `-b` wallet option so the user can explicitly select which blockchain client to use if multiple are available in the build. This will make adding new blockchain clients (such as #36) easier. Currently `electrum` is the default.
I also added a default esplora server url so the user doesn't need to specify one if selecting that client, which is how the electrum and compact_filters clients work.

The new wallet options looks like this (when all optional clients are enabled `--features esplora,compact_filters`):
```
bdk-cli-wallet 0.2.1-dev
Wallet mode

USAGE:
    bdk-cli wallet [FLAGS] [OPTIONS] --descriptor <DESCRIPTOR> <SUBCOMMAND>

FLAGS:
    -v, --verbose    Adds verbosity, returns PSBT in JSON format alongside serialized
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -n, --node <ADDRESS:PORT>...
            Compact filters blockchain client peer full node IP address:port [default: 127.0.0.1:18444]

    -b, --blockchain_client <BLOCKCHAIN_CLIENT>
            Blockchain client protocol [default: electrum]  [possible values: electrum, esplora, compact_filters]

    -c, --change_descriptor <CHANGE_DESCRIPTOR>        Sets the descriptor to use for internal addresses
        --conn_count <CONNECTIONS>
            Compact filters blockchain client number of parallel node connections [default: 4]

    -d, --descriptor <DESCRIPTOR>                      Sets the descriptor to use for the external addresses
        --esplora_concurrency <ESPLORA_CONCURRENCY>    Esplora blockchain client request concurrency [default: 4]
    -e, --esplora <ESPLORA_URL>
            Esplora blockchain client server url [default: https://blockstream.info/api/]

    -p, --proxy <PROXY_ADDRS:PORT>                     Blockchain client SOCKS5 proxy
    -r, --retries <PROXY_RETRIES>                      Blockchain client SOCKS5 proxy retries [default: 5]
    -t, --timeout <PROXY_TIMEOUT>                      Electrum blockchain client SOCKS5 proxy timeout
    -a, --proxy_auth <PROXY_USER:PASSWD>               Blockchain client SOCKS5 proxy credential
    -s, --server <SERVER:PORT>
            Electrum blockchain client server url [default: ssl://electrum.blockstream.info:60002]

    -k, --skip_blocks <SKIP_BLOCKS>
            Compact filters blockchain client skip initial blocks [default: 0]

    -w, --wallet <WALLET_NAME>                         Selects the wallet to use [default: main]
```

### Notes to the reviewers

In the docs for the `blockchain_client` option it displays the default `[electrum]` even if that feature is not enabled and all possible blockchain clients are displayed even if only some are enabled. I couldn't find a nice way to fix this, but the cli will throw an error if a blockchain client is selected that wasn't configured as a feature in the build.

I also simplified the CHANGELOG to focus on what a user would see as a change while using the bin.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,64 +6,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Project
-
-#### Added
-- `CliSubCommand::Compile` enum variant and `handle_compile_subcommand()` function
-
-#### Changed
-- Make repl and electrum default features
+- Update repl feature to not include electrum
 - Upgrade to `bdk` v0.7.x
-
-### `bdk-cli` bin
-
-#### Added
-- New top level command "Compile" which compiles a miniscript policy to an output descriptor
-- `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration 
-
-#### Changed
-- Remove unwraps while handling CLI commands
+- Add top level command "Compile" which compiles a miniscript policy to an output descriptor
+- Add `CompactFilterOpts` to `WalletOpts` to enable compact-filter blockchain configuration
+- Add `verbose` option to `WalletOpts` to display PSBTs also in JSON format
+- Add `blockchain_client` to `WalletOpts` to select blockchain client to use
 
 ## [0.2.0]
 
-### Project
-
-#### Added
 - Add support for `wasm`
-- `CliOpts` struct and `CliSubCommand` enum representing top level cli options and commands
-- `KeySubCommand` enum
-- `handle_key_subcommand` function
-
-#### Changed
 - Upgrade `bdk` to `0.4.0` and `bdk-macros` to `0.3.0`
-- Renamed `WalletOpt` struct to `WalletOpts`
-- `WalletSubCommand` enum split into `OfflineWalletSubCommand` and `OnlineWalletSubCommand`
-- Split `handle_wallet_subcommand` into two functions, `handle_offline_wallet_subcommand` and `handle_online_wallet_subcommand`
 - A wallet without a `Blockchain` is used when handling offline wallet sub-commands
-
-### `bdk-cli` bin
-
-#### Added
-- Top level commands "wallet", "key", and "repl"
-- "key" sub-commands to "generate" and "restore" a master private key
-- "key" sub-command to "derive" an extended public key from a master private key
+- Add top level commands "wallet", "key", and "repl"
+- Add "key" sub-commands to "generate" and "restore" a master private key
+- Add "key" sub-command to "derive" an extended public key from a master private key
 - "repl" command now has an "exit" sub-command
-
-#### Changed
 - "wallet" sub-commands and options must be proceeded by "wallet" command
 - "repl" command loop now includes both "wallet" and "key" sub-commands
 
 ## [0.1.0]
 
-### Project
-#### Added
 - Add CONTRIBUTING.md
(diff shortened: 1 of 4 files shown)


rajarshimaitra:
Thanks @notmandatory . This is a possible way to do it, but unfortunately, this also doesn't reduce our option arg namespace. So you will get the same conflicts with `-b` flag, as without it. Because `electrum` is on by default ( that turns on `ProxyOpts` too) even if we don't need it at all. 

I think the blockchain selection flag is kinda redundant because we are selecting blockchain backend at the build time itself with `--features` flag.

I have tried around a few things and it seems to me that the easiest thing to do to solve all of our problems, is if we drop the default blockchain, and just specify a backend each time we build `bdk-cli`. 

We can have the `repl` feature as default. 

Without any blockchain feature `bdk-cli` will just do the `repl` things. 

To have a full wallet we will specify a backend with `--feature` flag.

This will allow us to have the same arg option names for different blockchins. Because using feature guard removes the codes from the binary, so clap won't complain because for it those options don't exist.

And I think we can reasonably explain this in the usage docs too. 

I have tried this manually, and it produces nice compact `--help` doc also only with the blockchain that is enabled. like this

`--features rpc`
```
OPTIONS:
    -n, --rpc-node <ADDRESS:PORT>
            Sets the full node address for rpc connection [default: 127.0.0.1:18443]

    -c, --change_descriptor <CHANGE_DESCRIPTOR>    
            Sets the descriptor to use for internal addresses

    -d, --descriptor <DESCRIPTOR>                  
            Sets the descriptor to use for the external addresses

    -x, --skip-blocks <SKIP_BLOCKS>                
            Optionally skip initial `skip_blocks` blocks

    -A, --rpc-auth <USER:PASSWD>
            Sets the rpc authentication username:password [default: admin:password]

    -w, --wallet <WALLET_NAME>                     
            Selects the wallet to use [default: main]
```

`--features electrum`
```
OPTIONS:
    -c, --change_descriptor <CHANGE_DESCRIPTOR>    
            Sets the descriptor to use for internal addresses

    -d, --descriptor <DESCRIPTOR>                  
            Sets the descriptor to use for the external addresses

    -p, --proxy <PROXY_ADDRS:PORT>                 
            Sets the SOCKS5 proxy for Blockchain backend

    -r, --retries <PROXY_RETRIES>                  
            Sets the SOCKS5 proxy retries for the Electrum client [default: 5]

    -t, --timeout <PROXY_TIMEOUT>                  
            Sets the SOCKS5 proxy timeout for the Electrum client

    -a, --proxy-auth <PROXY_USER:PASSWD>           
            Sets the SOCKS5 proxy credential

    -s, --server <SERVER:PORT>
            Sets the Electrum server to use [default: ssl://electrum.blockstream.info:60002]

    -w, --wallet <WALLET_NAME>                     
            Selects the wallet to use [default: main]    
```

If this is something we wanna do, I can add it to my open PR. It's not a big change set.

notmandatory:
@rajarshimaitra OK if this doesn't solve the conflicting params issue I'll close it and open a new one with your one blockchain client at a time solution. I'll add some docs and a compile error if users try building with two blockchain client features enabled. I'd rather keep this change as a separate PR to keep it simple to review.

rajarshimaitra:
Yes that makes sense. Better to do it via separate PR.

PR #38, opened 2021-08-05
Allow enabling at most one blockchain client feature

notmandatory:
### Description

Allow at most one blockchain client feature be enabled at a time for builds. If no blockchain client feature is enabled then online wallet commands are disabled. This will simplify the options shown to the user and make adding new blockchain clients (such as #36) easier. ~~Electrum is still the default, to make a build with a different blockchain client the `--no-default-features` build option will need to be used.~~ No blockchain client is included in the `default` features, so if one is needed it must be specified with `--features`.  I also added a default esplora server url so the user doesn't need to specify one if selecting that client, which is how the electrum and compact_filters clients work.

### Notes to the reviewers

I changed the server option for both `electrum` and `esplora` to --server or -s since that now won't cause a conflict. I also simplified the CHANGELOG to focus on what a user would see as a change while using the bin.

I've also added a `build.rs` file to prevent more than one blockchain client feature from being enabled.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ compact_filters = ["bdk/compact_filters"]
 [[bin]]
 name = "bdk-cli"
 path = "src/bdk_cli.rs"
-required-features = ["repl", "electrum"]
+required-features = ["repl"]
 
 [package.metadata.docs.rs]
-features = ["esplora", "compiler"]
+features = ["compiler"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,9 @@ clap = { version = "2.33", optional = true }
 regex = {version = "1", optional = true }
 
 [features]
-default = ["repl", "electrum"]
-repl = ["bdk/key-value-db", "clap", "dirs-next", "env_logger", "regex", "rustyline"]
+default = ["cli", "repl"]
+cli = ["bdk/key-value-db", "clap", "dirs-next", "env_logger"]
+repl = ["regex", "rustyline"]
 electrum = ["bdk/electrum"]
 esplora = ["bdk/esplora"]
 compiler = ["bdk/compiler"]
@@ -39,7 +40,7 @@ compact_filters = ["bdk/compact_filters"]
 [[bin]]
 name = "bdk-cli"
 path = "src/bdk_cli.rs"
-required-features = ["repl"]
+required-features = ["cli"]
 
 [package.metadata.docs.rs]
 features = ["compiler"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,4 +43,4 @@ path = "src/bdk_cli.rs"
 required-features = ["cli"]
 
 [package.metadata.docs.rs]
-features = ["compiler"]
+features = ["compiler", "electrum"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ structopt = "^0.3"
 serde_json = { version = "^1.0" }
 log = "^0.4"
 base64 = "^0.11"
+zeroize = { version = "<1.4.0" }
 
 # Optional dependencies
 rustyline = { version = "6.0", optional = true }
--- a/README.md
+++ b/README.md
@@ -10,28 +10,31 @@ wallet tool based on the [bdk](https://github.com/bitcoindevkit/bdk) library.
 
 ## Install bdk-cli
 ### From source
-To install dev version of `bdk-cli` from local git repo:
+To install dev version of `bdk-cli` from local git repo with the `electrum` blockchain client enabled:
 
 ```shell
 cd <bdk-cli git repo directory>
-cargo install --path .
+cargo install --path . --features electrum
 bdk-cli help # to verify it worked
 ```
 
-By default the `electrum` client feature is enabled, to use the `esplora` or another blockchain backend
-client instead the default features must be disabled and the desired client feature enabled. Below
-is an example of how to enable the `esplora` blockchain client feature instead of `electrum`.
+If no blockchain client feature is enabled online wallet commands `sync` and `broadcast` will be 
+disabled. To enable these commands a blockchain client features such as `electrum` or another 
+blockchain backend feature must be enabled. Below is an example of how run the `bdk-cli` bin with 
+the `esplora` blockchain client feature.
 
 ```shell
-RUST_LOG=debug cargo run --no-default-features --features repl,esplora -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
+RUST_LOG=debug cargo run --features esplora -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
 ```
 
-If no blockchain client feature is enabled then online wallet commands are disabled.
+At most one blockchain feature can be enabled, available blockchain client features are:
+`electrum`, `esplora`, and `compact_filters`.
 
 ### From crates.io
-You can the install the binaries for the latest tag of `bdk-cli` directly from [crates.io](https://crates.io/crates/bdk-cli) like so:
+You can the install the binaries for the latest tag of `bdk-cli` with online wallet features 
+directly from [crates.io](https://crates.io/crates/bdk-cli) with a command as below:
 ```sh
-cargo install bdk-cli 
+cargo install bdk-cli --features `electrum`
 ```
 
 ### bdk-cli bin usage examples
@@ -46,7 +49,7 @@ cargo run
 To sync a wallet to the default electrum server:
 
 ```shell
-cargo run -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
+cargo run --features electrum -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
 ```
 
 To sync a wallet to Bitcoin Core node (assuming a regtest node at 127.0.0.1:18444) serving compact filters:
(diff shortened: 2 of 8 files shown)


rajarshimaitra (commented):
Thanks @notmandatory for the work. I just have a  few more comments. 

Is there any specific use of Repl mode here? https://github.com/bitcoindevkit/bdk-cli/blob/378b33aabcd2812ed431ed650a3f4418acc034d5/src/lib.rs#L258-L262
It just seems to be duplicating `WalletOpts` and no commands.

One problem here is using `--no-default-features` will turn off `repl` too, which is required dep for the binary, so it will not compile any binary. This has been the crux of the problem because cargo doesn't let us selectively disable features. So if we disable default, it disables everything in default.

I think the only way to solve this problem is by having a very minimal default with only the stuffs that we need to create the minimal possible binary. And then add stuff to it with features flag. This will also remove the use of `--no-default-feature` if we only want `esplora` but not `electrum`.

So I propose something like this:
```
[features]
default = ["repl"]
repl = ["bdk/key-value-db", "clap", "dirs-next", "env_logger", "regex", "rustyline"]
electrum = ["bdk/electrum"]
esplora = ["bdk/esplora"]
compiler = ["bdk/compiler"]
async-interface = ["bdk/async-interface"]
compact_filters = ["bdk/compact_filters"]

[[bin]]
name = "bdk-cli"
path = "src/bdk_cli.rs"
required-features = ["repl"]
```   
This will create the following builds
```
cargo build : build only repl stuffs, no backend
cargo build --features [backend] : build repl + [Backend] blockchain.
```
I don't think there's any adverse effect of not having blockchain by default. In fact, there can be usage where only some offline commands are required. In such a case we won't have to unnecessarily build a backend.

rajarshimaitra:
Also, I am not sure if we should silently ignore "no binary built" scenarios in CI. 

https://github.com/bitcoindevkit/bdk-cli/blob/378b33aabcd2812ed431ed650a3f4418acc034d5/.github/workflows/cont_integration.yml#L43

with current changes for example `cargo build --features esplora --no-default-features` will not build any binary, and CI won't complain.

notmandatory:
I think tests were passing because they only needed a lib build to run. But I agree it would be nice to not have to use the `--no-default-features` flag.  I did the following to try and address this:
1. split the `repl` feature into `cli` for the required dependencies to build the cli bin, and `repl` to enable the cli repl commands, then made the `cli` feature required to build the cli bin, but the `cli` and `repl` features on by default.
2. added cfg statements so someone could build the cli without the repl command (if they want to use --no-default-features)
3. update the CI to test with the default (no blockchain client) features, or each blockchain client, or the 'compiler' feature

src/bdk_cli.rs
```diff
@@ -104,29 +116,40 @@ fn new_online_wallet<D>(
...
         socks5: wallet_opts.proxy_opts.proxy.clone(),
         retry: wallet_opts.proxy_opts.retries,
         timeout: wallet_opts.electrum_opts.timeout,
     });
 
+    #[cfg(all(
+        feature = "esplora",
+        any(feature = "electrum", feature = "compact_filters")
+    ))]
+    compile_error!("Only one blockchain client feature can be enabled at a time.");
```
rajarshimaitra, line 137: It seems for every combination of `electrum`, `esplora` and `compact_filters`, this error is being duplicated with the one defined few lines below.

Instead, if we write it like this 
```
#[cfg(all(
        any( feature = "electrum", feature = "esplora", feature = "compact_filters"),
        any( feature = "electrum", feature = "esplora", feature = "compact_filters")
    ))]
```
this will activate for any combination of the above three. And we won't need to specify `compiler_error!` multiple times. 

Also for future extension, if we just add a new backend to the list above, that will taker care of multiple backend activation errors.

rajarshimaitra (commented):
Code review ACK. This simplifies the issue.

I had the following observation. It seems if we compile the library using `cargo build --features cli --no-default-features`, that produces a binary that is perfectly workable with non backend wallet. It has functionality like this 
```
$ ./target/debug/bdk-cli wallet --help
bdk-cli-wallet 0.2.1-dev
Wallet mode

USAGE:
    bdk-cli wallet [FLAGS] [OPTIONS] --descriptor <DESCRIPTOR> <SUBCOMMAND>

FLAGS:
    -v, --verbose    
            Adds verbosity, returns PSBT in JSON format alongside serialized

    -h, --help       
            Prints help information

    -V, --version    
            Prints version information


OPTIONS:
    -c, --change_descriptor <CHANGE_DESCRIPTOR>    
            Sets the descriptor to use for internal addresses

    -d, --descriptor <DESCRIPTOR>                  
            Sets the descriptor to use for the external addresses

    -w, --wallet <WALLET_NAME>                     
            Selects the wallet to use [default: main]


SUBCOMMANDS:
    bump_fee             Bumps the fees of an RBF transaction
    combine_psbt         Combines multiple PSBTs into one
    create_tx            Creates a new unsigned transaction
    extract_psbt         Extracts a raw transaction from a PSBT
    finalize_psbt        Finalizes a PSBT
    get_balance          Returns the current wallet balance
    get_new_address      Generates a new external address
    help                 Prints this message or the help of the given subcommand(s)
    list_transactions    Lists all the incoming and outgoing transactions of the wallet
    list_unspent         Lists the available spendable UTXOs
    policies             Returns the available spending policies for the descriptor
    public_descriptor    Returns the public version of the wallet's descriptor(s)
    sign                 Signs and tries to finalize a PSBT
```
So that begs the question I had before, what do we need the `repl` thing for? 

Also from here https://github.com/notmandatory/bdk-cli/blob/69b1d80bcd4715a3b78af4aafacebe2fc5ca2b88/src/bdk_cli.rs#L318-L331
it seems all a `repl` does is to perse existing commands and use them with either wallet or key functions. Maybe I am missing something, but I am not seeing any use of having the same methods called twice in different ways.

If two features `cli` and `repl` that does the same thing, it seems confusing to me.

rajarshimaitra:
The following structure and function call could use feature guard. They exist in binary compiled without a backend, but are never used.

https://github.com/notmandatory/bdk-cli/blob/69b1d80bcd4715a3b78af4aafacebe2fc5ca2b88/src/lib.rs#L892

https://github.com/notmandatory/bdk-cli/blob/69b1d80bcd4715a3b78af4aafacebe2fc5ca2b88/src/lib.rs#L638

src/bdk_cli.rs
notmandatory: Unfortunately the cfg all any any approach doesn't work, when I added this code:
```rust
#[cfg(all(
any( feature = "electrum", feature = "esplora", feature = "compact_filters"),
any( feature = "electrum", feature = "esplora", feature = "compact_filters")
))]
compile_error!("Only one blockchain client feature can be enabled at a time.");
```
I get the error even with only enable one blockchain client feature (ie. `cargo build --features esplora`). 

Looks like other's have had this same problem and are proposing a [Rust RFC](https://internals.rust-lang.org/t/pre-rfc-cargo-mutually-exclusive-features/13182/2) to allow a set of features to `provide` another feature in a mutually exclusive way.  

Until mutually exclusive features are possible I can fix this in a simplistic way in the `build.rs` file as below or just leave it as it.. What do you think?

```rust
fn main() {
    let electrum = env::var_os("CARGO_FEATURE_ELECTRUM").map(|_| "electrum".to_string());
    let esplora = env::var_os("CARGO_FEATURE_ESPLORA").map(|_| "esplora".to_string());
    let compact_filters = env::var_os("CARGO_FEATURE_COMPACT_FILTERS").map(|_| "compact_filters".to_string());

    let blockchain_features : Vec<String> = vec!(electrum, esplora, compact_filters)
        .iter()
        .map(|f| f.to_owned())
        .flatten()
        .collect();
    
    if blockchain_features.len() > 1 {
        panic!("At most one blockchain client feature can be enabled but these features were enabled: {:?}", blockchain_features)
    }
}
```

rajarshimaitra: Ah ok.. I thought it would work because my analyzer didn't throw. Yes there should be a way to handle mutually exclusive features in cargo itself.

I like the build script approach. It's clear and concise. Till it's available in cargo we can use a guard like above.

README.md
```diff
@@ -42,7 +49,7 @@ cargo run
 To sync a wallet to the default electrum server:
 
 ```shell
-cargo run -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
+cargo run --features electrum -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
```
rajarshimaitra, line 52: Few lines ago we mentioned how to install `bdk-cli` with `cargo`. 

Instructing to use `cargo run` here will build the local files instead of using the installed binary. So if someone is in their cloned repo, using cargo run they will actually build and then run the master branch instead of the release binary he just installed (this might be an unexpected behavior for the user, unless he knows cargo stuffs). 

Even worse if someone didn't clone the repo, `cargo run` will not run anything, even if they have `bdk-cli` binary installed (and we don't have that instruction anywhere).   

I think in the  "bdk-cli bin usage examples" section, we should just instruct to use `bdk-cli` binary instead of `crago run`, which will make the instruction simpler also.

This doesn't have to be fixed here. Just mentioned as it occurred to me.

src/lib.rs
```diff
@@ -431,7 +441,7 @@ pub struct ElectrumOpts {
     pub timeout: Option<u8>,
     /// Sets the Electrum server to use
     #[structopt(
-        name = "SERVER:PORT",
+        name = "ELECTRUM_URL",
```
rajarshimaitra, line 444: 5 lines below we have `pub electrum: String`. To keep it consistent with `esplora` maybe we can write `pub server: String`?

rajarshimaitra (commented):
tACK https://github.com/bitcoindevkit/bdk-cli/pull/38/commits/9bb1c608315603404a4db740e97dca14ac3334b7

With some minor nits.

Cargo.toml
```diff
@@ -38,7 +40,7 @@ compact_filters = ["bdk/compact_filters"]
 [[bin]]
 name = "bdk-cli"
 path = "src/bdk_cli.rs"
-required-features = ["repl", "electrum"]
+required-features = ["cli"]
```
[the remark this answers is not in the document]
notmandatory: I think we still need `cli` as required to force anyone building the bin without `repl` (using --no-default-features) to still include `cli`.

README.md
notmandatory: I agree we should probably update the README to focus on installing with different features, and then usage with an installed version instead of using `cargo run` since anyone making local changes should already know how to use cargo. But let's do that cleanup in a different PR.

src/lib.rs
notmandatory: Ah ya, not sure why I didn't change this one too, will fix.

Cargo.toml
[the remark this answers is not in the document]
rajarshimaitra: Ya makes sense.

rajarshimaitra:
ACK https://github.com/bitcoindevkit/bdk-cli/pull/38/commits/062542ae8957b2fef182588f7c786412545f228f

rajarshimaitra:
Code review with testing ACK https://github.com/bitcoindevkit/bdk-cli/pull/38/commits/4fd39b7734baa515867d00f6d710c4b2084de02d

rajarshimaitra (approved):
Code review with testing ACK https://github.com/bitcoindevkit/bdk-cli/pull/38/commits/4fd39b7734baa515867d00f6d710c4b2084de02d

notmandatory:
Thanks for the suggestions and review on this @rajarshimaitra, all ready to rebase your #36 PR.

PR #41, opened 2021-08-26
Update BDK version to 0.12.0

rajarshimaitra:
### Description

This PR updates the upstream BDK version to latest release (`v0.10.0`). 

### Notes to the reviewers

The latest release of BDK includes some breaking changes, some of which are handled here. A previously available function in BDK, `maintain_single_recipient()` is changed into `allow_shrinking()` which requires a scriptpubkey to be specified. So far I couldn't find any public API in BDK using which a scriptpubkey (or the transaction) can be retrieved from the wallet given a TxID. Marked the code with a `TODO`. It probably needs some additional API in BDK. Made an issue (https://github.com/bitcoindevkit/bdk/issues/427) to track this.

BDK `esplora` blockchain module has been divided into two types `using-ureq` and `using-reqwest`. `reqwest` is an async feature in BDK, which produces `Future` related errors in `bdk-cli` (https://github.com/bitcoindevkit/bdk-cli/pull/36#issue-702356592). Trying to solve that error creates more complexity (async related weridness). Thus decided to scrap `reqwest` and go with blocking `ureq` version, which I think is adequate for the usecases of `bdk-cli`.

Updated electrum cli args.     

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { version = "^0.7", default-features = false, features = ["all-keys"]}
+bdk = { version = "0.10.0", default-features = false, features = ["all-keys"]}
 bdk-macros = "^0.4"
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
@@ -32,7 +32,7 @@ default = ["cli", "repl"]
 cli = ["bdk/key-value-db", "clap", "dirs-next", "env_logger"]
 repl = ["regex", "rustyline"]
 electrum = ["bdk/electrum"]
-esplora = ["bdk/esplora"]
+esplora = ["bdk/use-esplora-ureq"]
 compiler = ["bdk/compiler"]
 async-interface = ["bdk/async-interface"]
 compact_filters = ["bdk/compact_filters"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { version = "0.10.0", default-features = false, features = ["all-keys"]}
+bdk = { version = "0.12.0", default-features = false, features = ["all-keys"]}
 bdk-macros = "^0.4"
 structopt = "^0.3"
 serde_json = { version = "^1.0" }
@@ -32,9 +32,10 @@ default = ["cli", "repl"]
 cli = ["bdk/key-value-db", "clap", "dirs-next", "env_logger"]
 repl = ["regex", "rustyline"]
 electrum = ["bdk/electrum"]
-esplora = ["bdk/use-esplora-ureq"]
+esplora = []
+esplora-ureq = ["esplora", "bdk/use-esplora-ureq"]
+esplora-reqwest = ["esplora", "bdk/use-esplora-reqwest"]
 compiler = ["bdk/compiler"]
-async-interface = ["bdk/async-interface"]
 compact_filters = ["bdk/compact_filters"]
 
 [[bin]]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk = { version = "^0.7", default-features = false, features = ["all-keys"]}
+bdk = { version = "0.10.0", default-features = false, features = ["all-keys"]}
```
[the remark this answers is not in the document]
rajarshimaitra: Good Idea. Better to enable `tls` in `ureq` if we use it. It seems like a general requirement to be able to connect https. Will open a PR if we agree..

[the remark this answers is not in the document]
notmandatory: True, maybe best of have `tls` feature enabled by default in `bdk`, but first I'd like to get @tcharding's input since he originally added the `ureq` dependency and may have an argument for leaving `tls` off by default.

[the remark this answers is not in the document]
tcharding: I have no problem with enabling TLS by default. As for why I disabled it, I can't remember exactly why, I probably just disabled default features during a debugging session and never turned TLS back on. Thanks for checking in with me and thanks for giving it the thought to realize we should enable it by default!

Cargo.toml
```diff
@@ -32,7 +32,7 @@ default = ["cli", "repl"]
 cli = ["bdk/key-value-db", "clap", "dirs-next", "env_logger"]
 repl = ["regex", "rustyline"]
 electrum = ["bdk/electrum"]
-esplora = ["bdk/esplora"]
+esplora = ["bdk/use-esplora-ureq"]
```
afilini, line 35: I'm trying to use this to test bitcoindevkit/bdk#429: to get bdk-cli to work in the browser we need a way to enable `bdk/use-esplora-reqwest`. I would suggest maybe adding another feature `esplora-reqwest` that we can use from wasm32

afilini: Or maybe even better: `async-esplora`

notmandatory: I think this PR may be blocked until @tcharding and @LLFourn  work out things out with https://github.com/bitcoindevkit/bdk/pull/433.

rajarshimaitra: Yes. I am not really sure how big the httpclient fix would be. If it takes much longer,  can we consider merging https://github.com/bitcoindevkit/bdk/pull/430 in the mean time? #36  is also pending on this one.

And I think that will fix @afilini 's issue too.

afilini: In the meantime I made this branch just to test the playground: https://github.com/afilini/bdk-cli/tree/update-with-reqwest

rajarshimaitra:
Added esplora-reqwest into bdk-cli. 

@notmandatory we have to use `bdk-master` for now, until the esplora-fix is released upstream. This can be changed in next bdk release. 

@afilini please check if this meets your requirement, or something more to be done?

notmandatory:
When `bdk` release `0.12.0` is out ~Sept-29 I think this one will be ready to merge. I've updated the title. :wink:

rajarshimaitra:
@notmandatory  just FYI, the rest of the open PRs builds on top of this. So this needs to go first before they can be merged.. :)

notmandatory:
I just made the [`bdk 0.12.0` release](https://crates.io/crates/bdk), so you're good to go on this one.

rajarshimaitra:
Rebased with bdk v0.12. 

FYI, I am not updating #42 and #36 right now. For simplicity of merging this needs to be merged first. Then the rest two can be rebased on top of updated master. I think that will simplify the review process as all these PRs are interdependent.

notmandatory (approved):
re-ACK f9a96062bce59b93e3449bc812c1a9d5d2ad684a

I manually tested and works fine as long as correct esplora server URL is used. Either README needs to be updated, defaults need to be changed for `testnet` or maybe something else. Will merge this now so other PRs can proceed and we can fix the esplora server URL issue in a subsequent PR.

PR #43, opened 2021-09-22
Use directly FromStr types in StructOpt

RCasatta:
With wrong values instead of getting:
[2021-09-21T13:25:13Z ERROR bdk_cli] Bip32(Base58(BadChecksum(1292914229, 1275556062)))

you get:
error: Invalid value for '--xprv <XPRV>': base58 encoding error: base58ck checksum 0x4c0770de does not match expected 0x4d104e35

#### All Submissions:

* [ ] I've signed all my commits
* [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -973,10 +973,10 @@ pub enum KeySubCommand {
     Derive {
         /// Extended private key to derive from
         #[structopt(name = "XPRV", short = "x", long = "xprv")]
-        xprv: String,
+        xprv: ExtendedPrivKey,
         /// Path to use to derive extended public key from extended private key
         #[structopt(name = "PATH", short = "p", long = "path")]
-        path: String,
+        path: DerivationPath,
     },
 }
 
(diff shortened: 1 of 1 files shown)


notmandatory:
LGTM, just need to fix `lib.rs` `test_key_derive()`, I think this is all it needs:
```rust
let key_generate_cmd = KeySubCommand::Derive {
    xprv: ExtendedPrivKey::from_str("tprv8ZgxMBicQKsPfQjJy8ge2cvBfDjLxJSkvNLVQiw7BQ5gTjKadG2rrcQB5zjcdaaUTz5EDNJaS77q4DzjqjogQBfMsaXFFNP3UqoBnwt2kyT").unwrap(),
    path: DerivationPath::from_str("m/84'/1'/0'/0").unwrap(),
};
```

notmandatory:
I also noticed that I'm getting the same output on a checksum error, even so I think this change is an improvement code-wise.
```shell
cargo run -- key derive --xprv "tprv8ZgxMBicQKsPfQjJy8ge2cvBfDjLxJSkvNLVQiw7BQ5gTjKadG2rrcQB5zjcdaaUTz5EDNJaS77q4DzjqjogQBfMsaXFFNP3UqoBnwt2ky" --path "m/84'/1'/0'/0" 
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/bdk-cli key derive --xprv tprv8ZgxMBicQKsPfQjJy8ge2cvBfDjLxJSkvNLVQiw7BQ5gTjKadG2rrcQB5zjcdaaUTz5EDNJaS77q4DzjqjogQBfMsaXFFNP3UqoBnwt2ky --path 'm/84'\''/1'\''/0'\''/0'`
error: Invalid value for '--xprv <XPRV>': base58 encoding error: base58ck checksum 0x125377bb does not match expected 0x5565d084
```

RCasatta:
> I also noticed that I'm getting the same output on a checksum error

The error message is improved because it displays which CLI parameter is wrong, which is very useful I think

notmandatory:
> The error message is improved because it displays which CLI parameter is wrong, which is very useful I think

Ah yes you're right, I wasn't in an old branch when I tested. This PR does produce a much better error message.

notmandatory (approved):
ACK 134cc06d9474b43118e36f6f1469f633ca473c12

PR #44, opened 2021-10-05
Small explora fixes

notmandatory:
### Description

* Fix default esplora server URL to `https://blockstream.info/testnet/api` to match default `bdk-cli` network which is also `testnet`. 
* Update README instructions and examples with features `esplora-ureq` and `esplora-reqest`.

### Notes to the reviewers

I think this is the simplest way to fix the esplora server URL to match the specified bitcoin network. If the user uses the option to change the bitcoin network then they should also expect to change the esplora (or other blockchain client) server url.

#### All Submissions:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,14 +183,14 @@ use bdk::{FeeRate, KeychainKind, Wallet};
 ///               },
 ///               #[cfg(feature = "esplora-ureq")]
 ///               esplora_opts: EsploraOpts {
-///                   server: "https://blockstream.info/api/".to_string(),
+///                   server: "https://blockstream.info/testnet/api/".to_string(),
 ///                   read_timeout: 5,
 ///                   write_timeout: 5,
 ///                   stop_gap: 10
 ///               },
 ///               #[cfg(feature = "esplora-reqwest")]
 ///               esplora_opts: EsploraOpts {
-///                   server: "https://blockstream.info/api/".to_string(),
+///                   server: "https://blockstream.info/testnet/api/".to_string(),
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1242,7 +1242,7 @@ mod test {
                     },
                     #[cfg(feature = "esplora")]
                     esplora_opts: EsploraOpts {
-                        server: "https://blockstream.info/api/".to_string(),
+                        server: "https://blockstream.info/testnet/api/".to_string(),
```
rajarshimaitra, line 1245: This seems redundant. Maybe we can remove this too? This is not activated in `electrum` tests and we don't have an `esplora` feature definition anymore.

src/lib.rs
```diff
@@ -1340,24 +1327,11 @@ mod test {
...
                         conc: 10,
                         stop_gap: 20
                     },
-                    #[cfg(feature = "compact_filters")]
-                    compactfilter_opts: CompactFilterOpts{
-                        address: vec!["127.0.0.1:18444".to_string()],
-                        skip_blocks: 0,
-                        conn_count: 4,
-                    },
-                    #[cfg(any(feature="compact_filters", feature="electrum", feature="esplora"))]
```
rajarshimaitra, line 1360: Agreed on removing redundant features.

rajarshimaitra (changes requested):
tACK https://github.com/bitcoindevkit/bdk-cli/pull/44/commits/cb17b5e57cbd44a7d61722e7f58528f548f97659

Verified that esplora sync is working with current changes.

PR #45, opened 2021-10-07
Replace `wallet bump_fee` command `--send_all` with new `--shrink` option

notmandatory:
### Description

Replace `wallet bump_fee` command `--send_all` with new `--shrink ADDRESS` option to reduce the output amount for the  specified address to increase RBF transaction fee.

### Notes to the reviewers

This new option is primarily needed when bumping the fee of a `send_all` transaction or any other transaction that doesn't have a change output that can be reduced to bump the fee on the new RBF transaction.

Steps I used to test
1. create wallet with two utxos 
2. create and broadcast send_all tx with rbf enabled and 1 sat/vbyte fee back to testnet faucet
3. create and broadcast bump_fee rbf for above tx with 7 sat/vbyte fee and shrinking the testnet faucet address

https://mempool.space/testnet/tx/e35892843875e084f39c5285ee6f522764210351f78cd89305a49efbf4bfea48

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Replace `wallet bump_fee` command `--send_all` with new `--shrink` option
+
 ## [0.3.0]
 
 - Add RPC backend support, after bdk v0.12.0 release
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -712,9 +712,9 @@ pub enum OfflineWalletSubCommand {
         /// TXID of the transaction to update
         #[structopt(name = "TXID", short = "txid", long = "txid")]
         txid: String,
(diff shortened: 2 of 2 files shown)


rajarshimaitra:
> For the purposed of this PR I think it's OK to allow bumping fees to external receivers even if there could be a missing case in the minimum fee logic, at least bdk requires the bump fee be greater than the current tx fee.

Sorry, I am not seeing how the minimum fee issue is relevant here. The issue is reducing external recipient's amount in RBF, which can be a potential bug. Min fee checking will always pass, if someone exploits this (the fee given will then be very high).  

https://github.com/bitcoindevkit/bdk/issues/256 are also fee amount related issues which is orthogonal to this.

notmandatory:
Although I see `bdk-cli` mostly as a test tool that should be able to create any valid bump fee transaction BDK and the bitcoin network will accept, I also see @rajarshimaitra's point that it should demonstrate wallet and BDK best practices.  

We can discuss at our team chat tomorrow and if no strong opinions against will merge. :-)

PR #48, opened 2021-10-18
adding the commands for proof-of-reserves using the separate repository

ulrichard:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,7 @@ dependencies = [
  "base64 0.11.0",
  "bdk",
  "bdk-macros",
+ "bdk-reserves",
  "clap",
  "dirs-next",
  "env_logger",
@@ -135,6 +136,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "bdk-reserves"
+version = "0.14.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "81e5def53134305f5c44dfd5b1b94609fc194d983d0b816e5ca7635d44a9d6dd"
+dependencies = [
+ "base64 0.11.0",
+ "bdk",
+ "bitcoinconsensus",
+ "log",
 ]
 
 [[package]]
@@ -200,6 +213,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "006cc91e1a1d99819bc5b8214be3555c1f0611b169f527a1fdc54ed1f2b745b0"
 dependencies = [
  "serde",
+]
+
+[[package]]
+name = "bitcoinconsensus"
+version = "0.19.0-3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1a8aa43b5cd02f856cb126a9af819e77b8910fdd74dd1407be649f2f5fe3a1b5"
+dependencies = [
+ "cc",
+ "libc",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ env_logger = { version = "0.7", optional = true }
 clap = { version = "2.33", optional = true }
 regex = { version = "1", optional = true }
 fd-lock = { version = "=3.0.0", optional = true }
+bdk-reserves = { version = "0.14.2", optional = true}
 
 [features]
 default = ["cli", "repl"]
@@ -39,6 +40,7 @@ esplora-reqwest = ["esplora", "bdk/use-esplora-reqwest"]
 compiler = ["bdk/compiler"]
 compact_filters = ["bdk/compact_filters"]
 rpc = ["bdk/rpc"]
+reserves = ["bdk-reserves"]
 
 [[bin]]
 name = "bdk-cli"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -122,6 +122,8 @@ use crate::OfflineWalletSubCommand::*;
     feature = "rpc"
 ))]
 use crate::OnlineWalletSubCommand::*;
+#[cfg(all(feature = "reserves", feature = "electrum"))]
+use bdk::bitcoin::blockdata::transaction::TxOut;
 use bdk::bitcoin::consensus::encode::{deserialize, serialize, serialize_hex};
 #[cfg(any(
     feature = "electrum",
@@ -134,6 +136,8 @@ use bdk::bitcoin::secp256k1::Secp256k1;
 use bdk::bitcoin::util::bip32::{DerivationPath, ExtendedPrivKey, KeySource};
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::{Address, Network, OutPoint, Script, Txid};
+#[cfg(feature = "reserves")]
+use bdk::blockchain::Capability;
 #[cfg(any(
     feature = "electrum",
     feature = "esplora",
@@ -145,6 +149,8 @@ use bdk::database::BatchDatabase;
 use bdk::descriptor::Segwitv0;
 #[cfg(feature = "compiler")]
 use bdk::descriptor::{Descriptor, Legacy, Miniscript};
+#[cfg(all(feature = "reserves", feature = "electrum"))]
+use bdk::electrum_client::{Client, ElectrumApi};
 use bdk::keys::bip39::{Language, Mnemonic, WordCount};
 use bdk::keys::DescriptorKey::Secret;
 use bdk::keys::KeyError::{InvalidNetwork, Message};
@@ -156,6 +162,16 @@ use bdk::wallet::AddressIndex;
 use bdk::Error;
 use bdk::SignOptions;
 use bdk::{FeeRate, KeychainKind, Wallet};
+#[cfg(all(feature = "reserves", feature = "electrum"))]
+use bdk_reserves::reserves::verify_proof;
+#[cfg(any(
+    feature = "electrum",
+    feature = "esplora",
+    feature = "compact_filters",
+    feature = "rpc"
+))]
+#[cfg(feature = "reserves")]
+use bdk_reserves::reserves::ProofOfReserves;
 
 /// Global options
 ///
@@ -307,6 +323,25 @@ pub enum CliSubCommand {
     Repl {
         #[structopt(flatten)]
         wallet_opts: WalletOpts,
+    },
+    /// Proof of reserves external sub-commands
+    #[cfg(all(feature = "reserves", feature = "electrum"))]
+    #[structopt(long_about = "Proof of reserves external verification")]
+    ExternalReserves {
+        /// Sets the challenge message with which the proof was produced
+        #[structopt(name = "MESSAGE", required = true, index = 1)]
+        message: String,
+        /// Sets the proof in form of a PSBT to verify
+        #[structopt(name = "PSBT", required = true, index = 2)]
+        psbt: String,
+        /// Sets the number of block confirmations for UTXOs to be considered.
+        #[structopt(name = "CONFIRMATIONS", required = true, index = 3)]
+        confirmations: usize,
+        /// Sets the addresses for which the proof was produced
+        #[structopt(name = "ADDRESSES", required = true, index = 4)]
+        addresses: Vec<String>,
+        #[structopt(flatten)]
+        electrum_opts: ElectrumOpts,
     },
 }
 
@@ -740,6 +775,9 @@ pub enum OfflineWalletSubCommand {
         /// Assume the blockchain has reached a specific height. This affects the transaction finalization, if there are timelocks in the descriptor
         #[structopt(name = "HEIGHT", long = "assume_height")]
         assume_height: Option<u32>,
+        /// Whether the signer should trust the witness_utxo, if the non_witness_utxo hasn’t been provided
+        #[structopt(name = "WITNESS", long = "trust_witness_utxo")]
+        trust_witness_utxo: Option<bool>,
     },
     /// Extracts a raw transaction from a PSBT
     ExtractPsbt {
@@ -755,6 +793,9 @@ pub enum OfflineWalletSubCommand {
         /// Assume the blockchain has reached a specific height
         #[structopt(name = "HEIGHT", long = "assume_height")]
         assume_height: Option<u32>,
+        /// Whether the signer should trust the witness_utxo, if the non_witness_utxo hasn’t been provided
+        #[structopt(name = "WITNESS", long = "trust_witness_utxo")]
+        trust_witness_utxo: Option<bool>,
     },
     /// Combines multiple PSBTs into one
     CombinePsbt {
@@ -808,6 +849,26 @@ pub enum OnlineWalletSubCommand {
             conflicts_with = "BASE64_PSBT"
         )]
         tx: Option<String>,
+    },
+    /// Produce a proof of reserves
+    #[cfg(feature = "reserves")]
+    ProduceProof {
+        /// Sets the message
+        #[structopt(name = "MESSAGE", long = "message")]
+        msg: String,
+    },
+    /// Verify a proof of reserves for our wallet
+    #[cfg(feature = "reserves")]
+    VerifyProof {
+        /// Sets the PSBT to verify
+        #[structopt(name = "BASE64_PSBT", long = "psbt")]
+        psbt: String,
+        /// Sets the message to verify
+        #[structopt(name = "MESSAGE", long = "message")]
+        msg: String,
+        /// Sets the number of block confirmations for UTXOs to be considered. If nothing is specified, 6 is used.
+        #[structopt(name = "CONFIRMATIONS", long = "confirmations", default_value = "6")]
+        confirmations: u32,
     },
 }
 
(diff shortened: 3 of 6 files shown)


ulrichard:
I just updated the implementation to better separate between internal and external proof validation. The internal validation with a wallet that can be built with a public key descriptor can be used with any blockchain connector. It is only the external validation with addresses that is only available with electrum. The latter one is the more important one for me, and I'm ok with it only being available with electrum for now.
I tested and validated the behavior with the following script. It also serves as a good documentation
[proof_of_reserves_roundtrip_blockfilters.sh.txt](https://github.com/bitcoindevkit/bdk-cli/files/7511678/proof_of_reserves_roundtrip_blockfilters.sh.txt)
.
Is there a standard way of integrating something like this in the documentation?

src/lib.rs
```diff
@@ -1219,6 +1301,70 @@ pub fn handle_compile_subcommand(
...
+    let client = Client::new(&electrum_opts.server).unwrap();
+
+    let outpoints_per_addr: Result<Vec<_>, _> = addresses
+        .iter()
+        .map(|address| {
+            let address = Address::from_str(&address).unwrap();
+            get_outpoints_for_address(address, &client)
+        })
+        .collect();
+    let outpoints_combined = outpoints_per_addr?
```
rajarshimaitra, line 1326: It's more clear to unwrap the the result in the previous step than before `iter()`.

Cargo.toml
```diff
@@ -38,6 +39,7 @@ esplora-reqwest = ["esplora", "bdk/use-esplora-reqwest"]
 compiler = ["bdk/compiler"]
 compact_filters = ["bdk/compact_filters"]
 rpc = ["bdk/rpc"]
+reserves = ["bdk/electrum", "bdk-reserves"]
```
rajarshimaitra, line 42: if you add the reserve feature like this, then inside the lib you don't have to specify both `reserve` and `electrum` feature gates. Only `reserve` would work as it would automatically open `electrum` stuffs.  
```suggestion
reserves = ["electrum", "bdk-reserves"]
```

src/lib.rs
```diff
@@ -1219,6 +1301,70 @@ pub fn handle_compile_subcommand(
     Ok(json!({"descriptor": descriptor.to_string()}))
 }
 
+/// Proof of reserves verification sub-command
+///
+/// Proof of reserves options are described in [`CliSubCommand::Reserves`].
```
rajarshimaitra, line 1306: nit:
```suggestion
/// Proof of reserves options are described in [`CliSubCommand::ExternalReserves`].
```

src/lib.rs
```diff
@@ -1219,6 +1301,70 @@ pub fn handle_compile_subcommand(
...
+#[cfg(all(feature = "reserves", feature = "electrum"))]
+pub fn handle_reserves_subcommand(
```
rajarshimaitra, line 1308: It would be better to name it `handle_ext_reserves_subcommand` or something. The internal reserves command are already handled by `OnlineWalletSubcommand`. So this might be confusing.

rajarshimaitra (changes requested):
untested Concept ACK. https://github.com/bitcoindevkit/bdk-cli/pull/48/commits/3d192ec924221aa4a5003f69bedc0daefe13525e

Thanks for working on this, it can be a really useful feature. And I like the current external trait impl on bdk::wallet, that makes the reserves crate really portable without putting it into bdk crate directly. But this could be eventually added to the BDK project as a util library.

I will try to test the whole thing in more detail. I am wondering what is stopping us from using other blockchains like core RPC. From your test script it seems its working fine with compact filters. Will look and report back.

Below are few minor nits in mean time.

src/lib.rs
ulrichard: I agree, that when reading the code, it would look more natural. 
But If I just move the "?" operator up, then the automatic type inference does not work as nicely, and neither the automatic abortion when hitting an item with Err. I experimented a while with that algo, and this is the cleanest that I found. It nicely makes use of moving the Err from any item out to the whole collection: https://doc.rust-lang.org/std/result/enum.Result.html#method.from_iter
Somehow I couldn't get that to work without a ton of type annotations when moving the ? operator to the first part.

Cargo.toml
ulrichard: Actually, reserves should not automatically pull in electrum. I missed this part in the recent separation.

ulrichard:
Yes, the plan for bdk-reserves was from the beginning to add it to the BDK project as a util library.

src/lib.rs
rajarshimaitra: You can use something like
```
    let outpoints_per_addr = addresses
        .iter()
        .map(|address| {
            let address = Address::from_str(&address)
                .map_err(|_| Error::Generic("Invalid address".to_string()))?;
            Ok(get_outpoints_for_address(address, &client)?)
        })
        .collect::<Result<Vec<Vec<_>>, Error>>()?;
```

rust automatically converts `Vec<Result<_>>` into `Result<Vec<_>>` by simple type declaration after `collect`. That can be used to unwrap the whole collection and also propagate the error.

src/lib.rs
```diff
@@ -1219,6 +1307,70 @@ pub fn handle_compile_subcommand(
...
+pub fn handle_ext_reserves_subcommand(
+    network: Network,
+    message: String,
+    addresses: Vec<String>,
+    psbt: String,
+    electrum_opts: ElectrumOpts,
+) -> Result<serde_json::Value, Error> {
+    let psbt = base64::decode(&psbt).unwrap();
+    let psbt: PartiallySignedTransaction = deserialize(&psbt).unwrap();
+    let client = Client::new(&electrum_opts.server).unwrap();
```
rajarshimaitra, line 1323: Although there are many places where we are using `unwrap`s currently in the lib, the goal is to handle them properly eventually. In that spirit it would be helpful if you can handle these unwraps. If there is no auto error conversion available for these functions, you can use the `Error::Generic()` variant to map them.

rajarshimaitra (changes requested):
initial tACK https://github.com/bitcoindevkit/bdk-cli/pull/48/commits/3cc81db190528031a1d67fe1409fd51393f54c58. I haven't run the test scripts, syncing my testnet node now. But there should not be an issue. Will play some more with it and report back if I find something odd.

Below are few more comments.

Cargo.toml
rajarshimaitra: Noted..

src/lib.rs
```diff
@@ -1936,4 +2103,268 @@ mod test {
...
+                        esplora: None,
+                        esplora_concurrency: 4,
+                    },
+                    #[cfg(any(feature = "compact_filters", feature = "electrum"))]
+                    proxy_opts: ProxyOpts {
+                        proxy: None,
+                        proxy_auth: None,
+                        retries: 5,
+                    },
+                },
```
rajarshimaitra, line 2151: Because we are feature gating this test at the very start, better not to include these redundant options.

src/lib.rs
```diff
@@ -1936,4 +2103,268 @@ mod test {
...
+                    esplora_opts: EsploraOpts {
+                        esplora: None,
+                        esplora_concurrency: 4,
+                    },
+                    #[cfg(any(feature = "compact_filters", feature = "electrum"))]
+                    proxy_opts: ProxyOpts {
+                        proxy: None,
+                        proxy_auth: None,
+                        retries: 5,
+                    },
```
rajarshimaitra, line 2207: Remove redundant.

src/lib.rs
```diff
@@ -1936,4 +2103,268 @@ mod test {
...
+                }),
+            },
+        };
+
+        assert_eq!(expected_cli_opts, cli_opts);
+    }
+
+    #[cfg(all(feature = "reserves", feature = "electrum"))]
+    #[test]
+    fn test_parse_verify_proof() {
```
rajarshimaitra, line 2163: I think a parsing test for "external descriptor" subcommand also needs to be added?

src/lib.rs
```diff
@@ -1936,4 +2103,268 @@ mod test {
...
+
+    /// Encodes a partially signed transaction as base64 and returns the  bytes of the resulting string.
+    #[cfg(all(feature = "reserves", feature = "electrum"))]
+    fn encode_tx(psbt: PartiallySignedTransaction) -> Vec<u8> {
+        let mut encoded = Vec::<u8>::new();
+        psbt.consensus_encode(&mut encoded).unwrap();
+        let tx = base64::encode(&encoded);
+
+        tx.as_bytes().to_vec()
+    }
```
rajarshimaitra, line 2227: just to make the namings a little clear
```suggestion
    fn encode_psbt(psbt: PartiallySignedTransaction) -> Vec<u8> {
        let mut encoded = Vec::<u8>::new();
        psbt.consensus_encode(&mut encoded).unwrap();
        let base64_psbt = base64::encode(&encoded);

        base64_psbt.as_bytes().to_vec()
    }
```

rajarshimaitra (changes requested):
reACK https://github.com/bitcoindevkit/bdk-cli/pull/48/commits/4a91e6347e60b2dcf5fedbb3250ec6c234dbbe22

Just few more small nits and I think its good to go.

src/lib.rs
```diff
@@ -1219,6 +1308,72 @@ pub fn handle_compile_subcommand(
...
+        })
+        .collect::<Result<Vec<Vec<_>>, Error>>()?;
+    let outpoints_combined = outpoints_per_addr
+        .iter()
+        .fold(Vec::new(), |mut outpoints, outs| {
+            outpoints.append(&mut outs.clone());
+            outpoints
+        });
+
+    let spendable = verify_proof(&psbt, &message, outpoints_combined, network).unwrap();
```
rajarshimaitra, line 1342: One unwrap missed..

rajarshimaitra (changes requested):

ulrichard:
> reACK [4a91e63](https://github.com/bitcoindevkit/bdk-cli/commit/4a91e6347e60b2dcf5fedbb3250ec6c234dbbe22)
> 
> Just few more small nits and I think its good to go.
> 
> Also I think the doc comments here
> 
> https://github.com/bitcoindevkit/bdk-cli/blob/f8f65f2bc56e46c34d17a0b6f8212e7f84933b5d/src/lib.rs#L192-L242
> and here
> 
> https://github.com/bitcoindevkit/bdk-cli/blob/f8f65f2bc56e46c34d17a0b6f8212e7f84933b5d/src/lib.rs#L362-L403
> needs to be updated to include the new flags and subcommands..

I don't see how the new sub commands would fit in these examples. They only list the wallet options. But reserves didn't add anything there. It only adds sub-commands.

rajarshimaitra (approved):
tACK https://github.com/bitcoindevkit/bdk-cli/pull/48/commits/4556d8e8743a60f8034ab76d651452932128ce93

notmandatory:
> I just updated the implementation to better separate between internal and external proof validation. The internal validation with a wallet (that can be built with a public key descriptor) can be used with any blockchain connector. It is only the external validation with addresses that is only available with electrum. The latter one is the more important one for me, and I'm ok with it only being available with electrum for now. I tested and validated the behavior with the following script. It also serves as a good documentation [proof_of_reserves_roundtrip_blockfilters.sh.txt](https://github.com/bitcoindevkit/bdk-cli/files/7511678/proof_of_reserves_roundtrip_blockfilters.sh.txt) Is there a standard way of integrating something like this in the documentation?

I think ultimately we can document this on the website docs in a chapter on `bdk-cli` and sub-chapter on the `reserves` feature. For now if you'd like you can do a web site blog post to describe the new functionality and how to use it which we can convert to the more formal docs later.

notmandatory (changes requested):
Tested and looks good. I have a few small requests to make the cli more consistent. 

Also I noticed that my proof of reserves validated even though one of the UTXOs wasn't confirmed on the (testnet) blockchain yet.  Is this a known issue?

notmandatory:
You also will need to rebase on master to pickup CI testing against our new stable rust version (1.56.1).

ulrichard:
> 



> Tested and looks good. I have a few small requests to make the cli more consistent.
> 
> Also I noticed that my proof of reserves validated even though one of the UTXOs wasn't confirmed on the (testnet) blockchain yet. Is this a known issue?

No, I was not aware of this. 
And this could be easily exploited, but also easily detected. 
I will add some checks and comments, also in the bdk-reserves lib.

src/lib.rs
```diff
@@ -809,6 +848,26 @@ pub enum OnlineWalletSubCommand {
...
+    #[cfg(feature = "reserves")]
+    VerifyProof {
+        /// Sets the PSBT to verify
+        #[structopt(name = "BASE64_PSBT", long = "psbt")]
+        psbt: String,
+        /// Sets the message to verify
+        #[structopt(name = "MESSAGE", long = "message")]
+        msg: String,
+        /// Sets the number of block confirmations for UTXOs to be considered. If nothing is specified, 6 is used.
+        #[structopt(name = "CONFIRMATIONS", long = "confirmations")]
```
notmandatory, line 868: I think a better way to specify the default for the cli is like this, it will also show in the help so you don't need to mention it in the above docs line:
```suggestion
        #[structopt(name = "CONFIRMATIONS", long = "confirmations", default_value = "6")]
```

src/lib.rs
```diff
@@ -809,6 +848,26 @@ pub enum OnlineWalletSubCommand {
...
+        confirmations: Option<u32>,
```
notmandatory, line 869: If you specify the `default_value` via struct opt then you don't need an Option here:
 ```suggestion
        confirmations: u32,
```

File: src/lib.rs [hunk 13]
```diff
@@ -1076,6 +1140,39 @@ where
...
+        #[cfg(feature = "reserves")]
+        VerifyProof {
+            psbt,
+            msg,
+            confirmations,
+        } => {
+            let psbt = base64::decode(&psbt).unwrap();
+            let psbt: PartiallySignedTransaction = deserialize(&psbt).unwrap();
+            let current_height = wallet.client().get_height()?;
+            let _max_confirmation_height = current_height - confirmations.unwrap_or(6);
```
notmandatory, line 1169: If default is specified via struct opt then this can be simplified a little (and remove hard coded value here):
```suggestion
            let _max_confirmation_height = current_height - confirmations;
```

ulrichard:
After adding the checks for block_height, many manual tests of the internal verify_proof() failed.
[proof_of_reserves_roundtrip.sh.txt](https://github.com/bitcoindevkit/bdk-cli/files/7642980/proof_of_reserves_roundtrip.sh.txt)
But the tests in https://github.com/weareseba/bdk-reserves/blob/main/tests/mempool.rs always succeed on my machine.
Is is possible that the transaction information is not fully stored in the database?

src/lib.rs
```diff
@@ -1076,6 +1140,39 @@ where
...
+            psbt,
+            msg,
+            confirmations,
+        } => {
+            let psbt = base64::decode(&psbt).unwrap();
+            let psbt: PartiallySignedTransaction = deserialize(&psbt).unwrap();
+            let current_height = wallet.client().get_height()?;
+            let max_confirmation_height = if confirmations == 0 { None } else { Some(current_height - confirmations) };
+
+            maybe_await!(wallet.sync(log_progress(), None))?;
```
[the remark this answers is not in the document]
notmandatory: If the user does a manual sync before calling the `verify_proof` command will it work? If so then I think that's ok and you don't need to add it in the code here.. Other command like `list_transactions`  etc. also expect the user to sync their wallet first.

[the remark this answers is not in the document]
ulrichard: No, user sync is not enough. You can verify that by commenting out this sync line, and then running the proof_of_reserves_roundtrip.sh.txt file that I attached to the comment above.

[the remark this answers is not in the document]
ulrichard: The interesting thing is that with user sync, but without this additional line wallet.list_unspent() works perfectly fine, but wallet.get_tx() does not.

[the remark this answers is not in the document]
ulrichard: Ah, I found the culprit. I had to rearrange the commands in proof_of_reserves_roundtrip.sh.txt so that the sync for the public descriptor wallet was performed right before their use. Before there were some calls with the private descriptor wallet in between. Somehow I assumed the default wallet name from descriptor (https://github.com/bitcoindevkit/bdk-cli/pull/57) was already in use.
With that I could remove the additional sync again.

notmandatory (approved):

notmandatory:
ACK  7e6a4c8
I ran through some manual testing using electrum and this looks ready to merge.

I would like to eventually change the integration tests to use `regtest` instead of `testnet` so the tests will be a little quicker and not break if the utxos in the test descriptor changes, but I think it's fine for now.

ulrichard:
Thanks for reviewing, testing and all the proposals for improvements @notmandatory and @rajarshimaitra 
I can't merge it myself, as I am not authorized.

rajarshimaitra:
@ulrichard Thanks for working on this. I will test the latest commits and merge it soon.

@notmandatory IIUC for regtest support we need this to work with RPC backend right?

notmandatory:
> @notmandatory IIUC for regtest support we need this to work with RPC backend right?

We can use the `bitcoind` and `electrsd` crates as we do in `bdk` to do automated regtest testing. But we haven't had any tests that needed them in this repo until the new `test_proof_of_reserves_*` tests were added.

Cargo.lock
```diff
@@ -137,6 +138,18 @@ dependencies = [
...
 
+[[package]]
+name = "bdk-reserves"
+version = "0.14.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "81e5def53134305f5c44dfd5b1b94609fc194d983d0b816e5ca7635d44a9d6dd"
+dependencies = [
+ "base64 0.11.0",
+ "bdk",
+ "bitcoinconsensus",
```
rajarshimaitra, line 149: `bdk` has a `verify` feature that uses `bitcoinconsensus`. Would there a be a way in future to use `bdk::verify` instead of `bitcoinconsensus` in the `bdk-reserves` dependency?

rajarshimaitra (approved):
ReACK https://github.com/bitcoindevkit/bdk-cli/pull/48/commits/7e6a4c848b0c7323ea12af074f38a42347a49fcc

@notmandatory yes that makes sense. Previously we didn't need such integrations tests, here, but now that we need one, maybe it would make sense to also add some generic wallet tests with regtest rpc too along with PoR tests? We should be able to do all the stuffs except the `ExternalReserves` thing with RPC. I will open a new issue to keep track of it.

@ulrichard I am merging this now. Below is one future improvement question.

PR #56, opened 2021-11-25
[bug] Fix RPC tests

rajarshimaitra:
### Description

There was an inconsistency in `rpc` tests. The `RpcOpts` was missing from the wallet struct. This wasn't caught before because we didn't have `rpc` tests in CI. Fixed that too.

Also removed redundant wallet options from feature gated tests.   

#### All Submissions:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1914,6 +1914,12 @@ mod test {
                         conn_count: 4,
                         skip_blocks: 0,
                     },
+                    #[cfg(feature = "rpc")]
+                    rpc_opts: RpcOpts {
+                        address: "127.0.0.1:18443".to_string(),
+                        auth: ("user".to_string(), "password".to_string()),
+                        skip_blocks: None,
+                    },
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1737,6 +1708,12 @@ mod test {
                         conn_count: 4,
                         skip_blocks: 0,
                     },
+                    #[cfg(feature = "rpc")]
```
notmandatory, line 1711: Since this whole tests has `#[cfg(feature = "rpc")]` I don't think you need it here. 
```suggestion
```

notmandatory (approved):
ACK 083a3e5af0e5a16f6bcd4d75bb4523aef5e01e66

notmandatory:
Looks like this one needs a rebase also before it can be merged.

rajarshimaitra:
Rebased.

notmandatory:
re-ACK  e3f3c3a

PR #58, opened 2021-11-27
Update codecov to grcov from Tarpauline

rajarshimaitra:
### Description

To match with existing codecov of BDK, Tarpaulin based codecov in bdk-cli is changed to grcov.

#### All Submissions:

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -3,25 +3,53 @@ on: [push]
 name: Code Coverage
 
 jobs:
-  tarpaulin-codecov:
-    name: Tarpaulin to codecov.io
+
+  codecov:
+    name: Code Coverage
     runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: '0'
+      RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off'
+      RUSTDOCFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off'
+
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Set default toolchain
-        run: rustup default nightly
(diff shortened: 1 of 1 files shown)


rajarshimaitra:
Although it seems we haven't been using the codecov reporting lately. There were no codecov reports for previous PRs.

@notmandatory is this intentional or something was off with Tarpaulin?

notmandatory:
Looks like a something needs to be done to fix the failure seen here: https://github.com/rajarshimaitra/bdk-cli/runs/4340753274?check_suite_focus=true

This project has a `rust-toolchain` file that's overriding the `nightly` toolchain, we either need to remove this file, or add to the `codecov` job: 
   ```shell
   rustup override set nightly
   ```

notmandatory:
> Although it seems we haven't been using the codecov reporting lately. There were no codecov reports for previous PRs.
> 
> @notmandatory is this intentional or something was off with Tarpaulin?

You mean no `codecov` reports for `bdk-cli` lately? I think they're being generated but not displayed inline, here's the latest one that was generated: https://app.codecov.io/gh/bitcoindevkit/bdk-cli/commit/b0c78e0aef181b2156c29549c3e27712166a4e49

rajarshimaitra:
> Looks like a something needs to be done to fix the failure seen here: 

I am not sure if we need the file. But for now I have set the override in CI.

>  I think they're being generated but not displayed inline,

Yes the coverage reports are generating fine, but they are not being displayed in PR page here. Also the same for bdk project. Maybe we need to add CodeCov App to github like this? https://docs.codecov.com/docs/how-to-create-a-github-app-for-codecov-enterprise

notmandatory (approved):
ACK 17c9329956a242db54d77680535c06d065d9c958

Also I think I've enabled the `Codecov` app for the `bitcoindevkit` org.. it should activate next time code is pushed to this repo.

notmandatory:
This one looks like it needs a rebase to get it to rerun the tests for rust 1.56.1 and then we can merge it.

rajarshimaitra:
Rebased.

PR #59, opened 2021-11-29
Update MSRV to current stable, 1.56.1

notmandatory:
### Description

Update MSRV to the current stable, 1.56.1 and update the `cargo.lock` file.  Also removed the `rust-toolchain` file.

### Notes to the reviewers

I removed the `rust-toolchain` file because I don't think we need to lock users into a particular version (or force them to override it), and this file seem to primarily be used for projects using a specific nightly release, see: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file.

I also had to pin the `fd-lock` dependency because of a recent update to that crate to use the [`rustdoc::missing_doc_code_examples` feature which is only supported by `nightly`](https://doc.rust-lang.org/rustdoc/lints.html#missing_doc_code_examples) :disappointed: . 

#### All Submissions:


notmandatory:
Before this is merged we need to change the github expected checks from `1.53.0` to `1.56.1`.

notmandatory:
undoing the pinned fd-lock with #64

Issue #55, opened 2021-11-23
Add options to start regtest bitcoind and electrsd


notmandatory:
For users who are manually testing `bdk` with `bdk-cli` in regtest mode, add a new `regtest-net` cargo feature that will automatically start and stop a `bitcoind` and if needed an `electrsd` daemon with `electrum` or `esplora` support. The regtest net daemons will only be started if the `--network regtest` option is enabled. 

The feature should work similarly to the `bdk` integration tests but store node data in the users `~/.bdk-bitcoin` directory and not a random temp directory, and use standard regtest network ports.  

Using this feature would look something like this:

```shell
# start regtest `bitcoind` and `electrs` and stop daemons when repl (or wallet command) finishes
cargo run --features regtest-net,electrum -- --network regtest repl --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" 
```
By enabling the `regtest-net` feature when running or installing `bdk-cli` a localhost `bitcoind` and (if `electrum` or `esplora` features enabled) `electrsd` daemon would be started and stopped for each command. This will be most convenient when running multiple commands via the `repl` shell but still usable (but slow) for single `wallet` commands. 

The [`bitcoind`](https://github.com/RCasatta/bitcoind) and [`electrsd`](https://github.com/rcasatta/electrsd) crates provide most of what is needed to download, start, and stop the needed daemons. This will only work on linux and macos systems.

It may also be useful to add additional `wallet` commands to get a new address from `bitcoind` or trigger `bitcoind` to send regtest coins to the `bdk-cli` wallet.

rajarshimaitra:
This would be really helpful to make bdk-cli testing process more seamless..

PR #65, opened 2021-12-20
Auto deploy regtest nodes in background

rajarshimaitra:
### Description

Fixes https://github.com/bitcoindevkit/bdk-cli/issues/55. I decided to go for one feature flag for each backends as they made dependency management easier. Because different backend will have different dependency, and we don't wanna have redundant ones. Also it makes the commands a little shorter.
   
### Notes to the reviewers

Right now its not that useful because the backend APIs are not exposed. So I can't really do much other than sync/send in the command line. (unlike testnet we cannot send coins to bdk from outside of local regtest node). So to make it useful in repl mode as described in the issue, we also have to expose the blockchain apis in bdk-cli command in some way (at least the basic stuffs, address, send, generate). Looking for suggestions on this.

usage looks like this: 
```
cargo run --features regtest-rpc -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
{}
```

Although I am having trouble connecting to the esplora end of electrs. Not sure whats the issue, will have to look more. 

#### All Submissions:

#### New Features:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -19,13 +19,19 @@ jobs:
           - esplora-reqwest
           - compiler
           - compact_filters
+          - rpc
           - reserves
           - reserves,electrum
           - reserves,esplora-ureq
           - reserves,compact_filters
           - reserves,rpc
-          - rpc
           - electrum,verify
+          # regtest-* features are experimental and not fully usable
+          - regtest-bitcoin
+          - regtest-electrum
+          - regtest-esplora-ureq
+          - regtest-esplora-reqwest
+
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Re-license to dual MIT and Apache 2.0 and update project name to "Bitcoin Dev Kit"
 - Update to bdk and bdk-reserves to `0.17.0`
 - Add 'verify' feature flag which enables transaction verification against consensus rules during sync.
+- Make `regtest` the default network.
+- Add experimental `regtest-*` features to automatically deploy local regtest nodes
+(bitcoind, and electrs) while running cli commands.
 
 ## [0.4.0]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -47,6 +53,12 @@ name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
+name = "base-x"
+version = "0.2.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4521f3e3d031370679b3b140beb36dfe4801b09ac77e30c61941f97df3ef28b"
 
 [[package]]
 name = "base64"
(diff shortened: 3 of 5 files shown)


rajarshimaitra:
Fixed merge conflict with some enhancements.

rajarshimaitra:
Updated the commit history for better structuring.

notmandatory:
@rajarshimaitra Looks like this one needs a rebase or if you're tied up I can do the rebase but would have to rewrite your commits. Once that's done I'll do some testing and see if I can figure out how to expose the core rpc in repl mode.

rajarshimaitra:
Rebased and resolved conflicts..

notmandatory:
I spent some time today testing and reviewing and this is a great start! There are still the two (related) issues you found:

- [ ] 1. no way to control the bitcoind node (ie. create new blocks, get regtest coins)
- [ ] 2. sync doesn't work for esplora

Plus one I'd like to fix: 

- [ ] 3. too many features, I propose we use one `regtest` feature

The fix for 3 is easy, my proposal is here: https://github.com/notmandatory/bdk-cli/tree/regtest-node-feat

I believe issue 2 is because `electrsd` won't sync until at least one block is created by `bitcoind`; so if we fix 1 we should also fix 2. My idea for fixing 1 is to change the `bitcoind` lib to allow us to set a reusable (not randomly generated tmp) directory for the `bitcoind` data dir; and maybe the same for `electrsd`. Then we can add some simple RPC commands to `bdk-cli` to update the `bitcoind` (and `electrsd`) database and that data will persist for future invocations of `bdk-cli`. 

But for now if we only fix 3 I think it's OK to merge this PR, even if a little broken, as long as we add the comment that the `regtest` feature is experimental. Then we have more time to to play with and get merged any PRs to `bitcoind` (and `electrsd`) that we need to have it all working end to end.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add 'reserve' feature to enable proof of reserve
 - If no wallet name is provided, derive one from the descriptor instead of using "main"
 - Add optional cookie authentication for rpc backend
+- Add `regtest-*` features to automatically deploy local regtest nodes
```
notmandatory, line 13:
```suggestion
- Add experimental `regtest-*` features to automatically deploy local regtest nodes
```

.github/workflows/cont_integration.yml
```diff
@@ -19,11 +19,16 @@ jobs:
           - esplora-reqwest
           - compiler
           - compact_filters
+          - rpc
           - reserves
           - reserves,electrum
           - reserves,esplora-ureq
           - reserves,compact_filters
-          - rpc
+          - regtest-bitcoin
```
notmandatory, line 27:
```suggestion
          # regtest-* features are experimental and not fully usable
          - regtest-bitcoin
```

notmandatory (changes requested):
Add some little experimental warnings and this is good to go.

rajarshimaitra:
Thanks @notmandatory .. Updated as per suggested and rebased..

notmandatory (approved):

src/lib.rs
```diff
@@ -260,7 +260,7 @@ pub struct CliOpts {
         name = "NETWORK",
         short = "n",
         long = "network",
-        default_value = "testnet"
+        default_value = "regtest"
```
notmandatory, line 263: I didn't see this change before, sorry but I don't think we should make "regtest" the default network. Most casual users will expect "testnet" since that's how existing docs and blogs are written.

notmandatory (changes requested):
Also need to do a cargo fmt.

src/lib.rs
rajarshimaitra: Ya I thought about that too. But to me it occurred that bdk-cli is not for casual user but for devs trying out their own bdk wallet setup. I find myself setting the network to regtest all the time. Because casual testing is harder with testnet. Also the `BitcoinD` backend crate defaults to regtest. It would be messy if we try to make it testnet and then delete temporary datadir after each execution.. Testnet takes a long time to sync. Which is not suitable for casual testing. 

I feel testnet is useful when devs already have a prototype working and want to simulate mainnet type behavior without real money.

So does this make a good justification to make the default to regtest? We can change the docs and other instructions at other places..

notmandatory: How about for now we leave testnet.. but once this easy regtest node setup is ready for prime time we make regtest the default? That will also give us time to update the docs before it takes effect.

rajarshimaitra: Yes makes sense..

rajarshimaitra:
- Reverted back default network to `testnet`
 - Fixed test failure

notmandatory (approved):
ACK 930b0c48fef484e5b586346ca295358a3d30bb5e

Issue #67, opened 2021-12-29
Add Cookie authetication


rajarshimaitra:
bdk-cli currently uses rpc "user:password" authentication.

We can also allow it to connect to rpc via cookie file. Although looking for opinions on how useful it can be.

PR #68, opened 2022-01-08
Add cookie authentication option for rpc

rajarshimaitra:
### Description

Fixes https://github.com/bitcoindevkit/bdk-cli/issues/67

This PR adds an optional `cookie` arg to pass on cookie file if user needs to use that instead of basic user:password based authentication.

bitcoin core will only generate a `.cookie` file if rpc username passwords are not set in `bitcoin.conf`. Thus cookie file doesn't have a default. User should only provide it if they know a cookie file exists, or else "File not found" error is returned.     

#### All Submissions:

#### New Features:


notmandatory (approved):
utACK 82f7ff28458010a133403d13c4fd7bfa8146e9fb.

But needs another rebase. After than I'll give it a final test and should be good to merge.

rajarshimaitra:
Rebased..

notmandatory (approved):
ACK 1ab2b0ffc780defa913ec9ce2dc021b355c2aeb3

PR #71, opened 2022-02-18
Add key-value-db and sqlite-db features, separate wallet directories

notmandatory:
### Description

- Add distinct `key-value-db` and `sqlite-db` features, keep default as `key-value-db`

- Put cached wallet data in separate directories: `~/.bdk-bitcoin/<wallet_name>`
- Put compact filter data in `<wallet_name>/compact_filters`
- Depending on the db used put cached wallet data in: `<wallet_name>/wallet.sled/` or `<wallet_name>/wallet.sqlite`

### Notes to the reviewers

This change will help test BDK with different databases, in particular for manually testing DB migrations such as in https://github.com/bitcoindevkit/bdk/pull/502.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/bdk_cli.rs
+++ b/src/bdk_cli.rs
@@ -91,7 +91,7 @@ enum ReplSubCommand {
 }
 
 /// prepare bdk_cli home and wallet directory
-fn prepare_home_wallet_dir(wallet_name: &String) -> Result<PathBuf, Error> {
+fn prepare_home_wallet_dir(wallet_name: &str) -> Result<PathBuf, Error> {
     let mut dir = PathBuf::new();
     dir.push(
         &dirs_next::home_dir().ok_or_else(|| Error::Generic("home dir not found".to_string()))?,
@@ -114,7 +114,7 @@ fn prepare_home_wallet_dir(wallet_name: &String) -> Result<PathBuf, Error> {
 }
 
 /// Prepare wallet database directory
-fn prepare_wallet_db_dir(wallet_name: &String) -> Result<PathBuf, Error> {
+fn prepare_wallet_db_dir(wallet_name: &str) -> Result<PathBuf, Error> {
     let mut db_dir = prepare_home_wallet_dir(wallet_name)?;
 
     #[cfg(feature = "key-value-db")]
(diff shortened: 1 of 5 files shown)


notmandatory:
@rajarshimaitra can you take a look at this one?

src/bdk_cli.rs
```diff
@@ -112,25 +115,75 @@ fn prepare_home_dir() -> Result<PathBuf, Error> {
...
+        info!("Creating database directory {}", db_dir.as_path().display());
+        fs::create_dir(&db_dir).map_err(|e| Error::Generic(e.to_string()))?;
+    }
+
+    Ok(db_dir)
+}
+
+/// Prepare blockchain data directory (for compact filters)
+#[cfg(feature = "compact_filters")]
+fn prepare_bc_dir(wallet_name: &str) -> Result<PathBuf, Error> {
```
rajarshimaitra, line 149: Just a question: something like this could also be used for auto regtest node datadir right?

rajarshimaitra (commented):
Review ACK https://github.com/bitcoindevkit/bdk-cli/pull/71/commits/2e3801151afeb81767605b674c643e8b0a42b454

Looks good for now.. Will review again when additive features stuff is done..

src/bdk_cli.rs
notmandatory: Yes! I'd like to put the auto regtest data someplace like:
* `~/.bdk-bitcoin/<wallet_name>/regtest/bitcoind/` and
* `~/.bdk-bitcoin/<wallet_name>/regest/electrsd/`.

notmandatory:
@rajarshimaitra I changed my mind again on the additive features thing. I still think it'd be good to try but will big a biggish change (if it's even possible) for bdk-cli and probably not worth the effort right now.

This PR is now rebased and should be ready to go.

src/bdk_cli.rs
```diff
@@ -100,25 +103,75 @@ fn prepare_home_dir() -> Result<PathBuf, Error> {
...
+    db_dir.push("wallet.sled");
+
+    #[cfg(feature = "sqlite-db")]
+    db_dir.push("wallet.sqlite");
+
+    #[cfg(not(feature = "sqlite-db"))]
+    if !db_dir.exists() {
+        info!("Creating database directory {}", db_dir.as_path().display());
+        fs::create_dir(&db_dir).map_err(|e| Error::Generic(e.to_string()))?;
+    }
```
rajarshimaitra, line 130: I did not understand the reason of feature guard here. For both sled and sqlite we would want to check if the dir exists right??

rajarshimaitra (commented):
Ack 00454be42be22907d4d9c959483e79dc662de71c

With just one question..

src/bdk_cli.rs
notmandatory: The reason is that sqlite creates only a single file and doesn't need a directory to exist to put it's files in.

PR #90, opened 2022-05-11
Disable regex tests when repl feature not enabled

notmandatory:
### Description

I noticed that when I disable default features and don't enable the `repl` feature that the regex tests fail because they need the regex dependency that the `repl` feature brings in. This PR is to simply disable the regex related test code if the `repl` feature isn't enabled.

#### All Submissions:

#### New Features:

#### Bugfixes:


rajarshimaitra (approved):
ACK 953c097871d34d2606d55294313b1addc579dfa4

Repro'd the bug using
```
cargo test --features=cli,key-value-db --no-default-features
```

PR #104, opened 2022-06-27
Update crate documentation

rajarshimaitra:
### Description

After #99 the previous documentation have been removed and new docs as per `structopts` documentation. This PR adds more documentation across the crate.. 

This PR is above #102 , to accommodate all the further refactoring changes. 

The Readme About section have been updated with more details.. Readme format made aligned with the BDK project itself..

The Readme file is used itself as the crate level documentation in docs.rs too..   

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #116, opened 2022-09-02
Create wallet dir only when necessary

danielabrozzoni:
### Description

Making sure to create the wallet dir only when necessary (database is either sqlite or key-value)

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -164,6 +164,11 @@ pub(crate) fn prepare_home_dir(home_path: Option<PathBuf>) -> Result<PathBuf, Er
 }
 
 /// Prepare bdk_cli wallet directory.
+#[cfg(any(
+    feature = "key-value-db",
+    feature = "sqlite-db",
+    feature = "compact_filters"
+))]
 fn prepare_wallet_dir(wallet_name: &str, home_path: &Path) -> Result<PathBuf, Error> {
     let mut dir = home_path.to_owned();
 
@@ -178,6 +183,7 @@ fn prepare_wallet_dir(wallet_name: &str, home_path: &Path) -> Result<PathBuf, Er
 }
 
 /// Prepare wallet database directory.
+#[cfg(any(feature = "key-value-db", feature = "sqlite-db",))]
 fn prepare_wallet_db_dir(wallet_name: &str, home_path: &Path) -> Result<PathBuf, Error> {
     let mut db_dir = prepare_wallet_dir(wallet_name, home_path)?;
 
@@ -248,12 +254,14 @@ pub(crate) fn prepare_electrum_datadir(home_path: &Path) -> Result<PathBuf, Erro
     Ok(dir)
 }
 
+#[allow(unused_variables)]
 /// Open the wallet database.
 pub(crate) fn open_database(
     wallet_opts: &WalletOpts,
     home_path: &Path,
 ) -> Result<AnyDatabase, Error> {
     let wallet_name = wallet_opts.wallet.as_ref().expect("wallet name");
+    #[cfg(any(feature = "key-value-db", feature = "sqlite-db",))]
     let database_path = prepare_wallet_db_dir(wallet_name, home_path)?;
 
     #[cfg(feature = "key-value-db")]


rajarshimaitra:
@danielabrozzoni I have modified this part of the code in #102 .. We needed to streamline the directory creation for many directories (includeing bitcoind and electrsd dirs along with wallet files) and current function flow wasn't ideal for it.. 

I don't know whats the good way to go, but all the new PRs in these locations are going to heavily conflict with #102.. Its always this problem of having big refactoring PR open for a long time. All downstream works gets conflicted.. 

Maybe we should push in #102 and #104 before making further updates? The resultant structure of the library might also guide approach for the following for wasm changes.. The current state of the library is an "intermediate" one..

danielabrozzoni:
I understand if you don't want to rebase your PR on top of this one, but at least check that this same logic of "let's avoid creating the data directory if we don't need it" (which is a prerequisite for wasm) is still doable after #102, otherwise I have to re-refactor the whole code.

danielabrozzoni:
Also, how long will it take for #102 and #104? They are both huge, are they going to be merged soon?

rajarshimaitra:
> I understand if you don't want to rebase your PR on top of this one, but at least check that this same logic of "let's avoid creating the data directory if we don't need it" (which is a prerequisite for wasm) is still doable after https://github.com/bitcoindevkit/bdk-cli/pull/102, otherwise I have to re-refactor the whole code.

The logic looks good to me.. 102 is meant to make wasm works easier, the `handle_command` function there only takes a `CliOpts`, so the wasm code should be easier to write, as you don't need to pass in a backend..  This PR can be done over 102 too, but probably over a different function. 

> Also, how long will it take for https://github.com/bitcoindevkit/bdk-cli/pull/102 and https://github.com/bitcoindevkit/bdk-cli/pull/104? They are both huge, are they going to be merged soon?

The code is ready.. No new modification would be required.. If small fixes remains can be done in future PRs.. We are just waiting for bdk 22.. After that it should be good to go..

notmandatory:
> But there are still no ACKs, and the review process will take quite some time, so it will probably be merged in a couple of weeks? Does it make sense to wait weeks before merging this one? As you wish, for me it's fine rebasing this one, it just doesn't seem worth it to make it wait so long...

I just now added an ACK on #102 with a little explanation, I think it can be made ready to merge shortly after bdk 0.22 is released.  My preference would be to then make the next bdk-cli release. Possibly with this PR also if it's not a big deal to rebase.

notmandatory (approved):
tACK 74f65534987902450c72a86efbbbb9a4a5c7f4c3

I used below command to confirm wallet worked and no dirs created as expected. 
```
cargo run --no-default-features --features esplora-ureq,repl -- repl --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)"
```

rajarshimaitra:
> Does it make sense to wait weeks before merging this one?

I am ok with merging this one.. But its getting really difficult to rebase 102 with more inb4 changes.. It could even require rewriting some changes totally, depending upon where it is being made.. 

For this one its straight forward so its fine to go in.. Probably the feature gate will moved onto a different function, but I can handle that at rebase.. But more intrusive changes at this stage in the crate would get difficult.. 

I am expecting 102 to go in as soon as the next bdk 22 release..

danielabrozzoni:
I thought it would take more to get #102 merged, if it's merged shortly after 0.22 then I don't mind waiting and rebasing :)

-------- Original Message --------

danielabrozzoni:
After #102 this is quite difficult to do, but rebasing #117 I noticed that we might not need it for wasm...

danielabrozzoni:
Nevermind, I ended up needing it for #117. This is now ready for review :)

src/utils.rs
```diff
@@ -252,7 +258,9 @@ pub(crate) fn open_database(
     wallet_opts: &WalletOpts,
     home_path: &Path,
 ) -> Result<AnyDatabase, Error> {
+    #![allow(unused_variables)]
     let wallet_name = wallet_opts.wallet.as_ref().expect("wallet name");
+    #[cfg(any(feature = "key-value-db", feature = "sqlite-db",))]
     let database_path = prepare_wallet_db_dir(wallet_name, home_path)?;
```
rajarshimaitra, line 264: The `home_path` is also unused here in case of memory db. So thats triggering another unused warning.. Why not just mark the whole function with `#[allow(unused_variables)]`. That takes care of all the unused..

rajarshimaitra (commented):
tACK a4f58dcd8726beff96c4d527fad25672304df57a

Just one nit..

src/utils.rs
rajarshimaitra: Oh because you are using an inner attribute.. Just do outer attribute like  `#[allow(unused_variable)]` without the `!` mark.. That should work..

danielabrozzoni: Oh that's right, thanks! I just updated the code :)

PR #117, opened 2022-09-05
Update bdk cli for wasm

danielabrozzoni:
### Description

This PR adds a module to bdk-cli that exposes some structures/functions to WASM. You can see it in use here: https://github.com/bitcoindevkit/bitcoindevkit.org/pull/118

### Notes to the reviewers

Based on #116

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -55,6 +55,41 @@ jobs:
       - name: Test
         run: cargo test --no-default-features --features repl,${{ matrix.features }}
 
+  wasm-build:
+    name: Build WASM
+    runs-on: ubuntu-latest
+    env:
+      CC: clang-10
+      CFLAGS: -I/usr/include
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Generate cache key
+        run: echo "Build WASM" | tee .cache_key
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('.cache_key') }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+        # Install a recent version of clang that supports wasm32
+      - run: wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add - || exit 1
+      - run: sudo apt-add-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main" || exit 1
+      - run: sudo apt-get update || exit 1
+      - run: sudo apt-get install -y libclang-common-10-dev clang-10 libc6-dev-i386 || exit 1
+      - name: Set default toolchain
+        run: rustup default 1.56.1 # MSRV
+      - name: Set profile
+        run: rustup set profile minimal
+      - name: Add target wasm32
+        run: rustup target add wasm32-unknown-unknown
+      - name: Update toolchain
+        run: rustup update
+      - name: Build
+        run: cargo build --target wasm32-unknown-unknown --no-default-features --features esplora-reqwest,async-interface,compiler
+
   fmt:
     name: Rust fmt
     runs-on: ubuntu-latest
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -1,0 +1,200 @@
+use crate::commands::*;
+use crate::handlers::*;
+use crate::nodes::Nodes;
+use crate::utils::*;
+use bdk::*;
+
+use bitcoin::*;
+
+use bdk::blockchain::AnyBlockchain;
+use bdk::database::AnyDatabase;
+use js_sys::Promise;
+use regex::Regex;
+use std::error::Error;
+use std::ops::Deref;
+use std::path::PathBuf;
+use std::rc::Rc;
+use std::str::FromStr;
+use structopt::StructOpt;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::future_to_promise;
+
+#[cfg(feature = "compiler")]
+use bdk::keys::{GeneratableDefaultOptions, GeneratedKey};
+#[cfg(feature = "compiler")]
+use bdk::miniscript::{self, policy::Concrete, Descriptor, TranslatePk};
+#[cfg(feature = "compiler")]
+use serde::Deserialize;
+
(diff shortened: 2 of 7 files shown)


src/wasm.rs
```diff
@@ -0,0 +1,92 @@
...
+                .into_iter()
+                .map(|a| a.as_string().expect("Invalid type"));
+            let repl_subcommand =
+                ReplSubCommand::from_iter_safe(command).map_err(|e| e.to_string())?;
+            log::debug!("repl_subcommand = {:?}", repl_subcommand);
+
+            let result = match repl_subcommand {
+                ReplSubCommand::OnlineWalletSubCommand(online_subcommand) => {
+                    let blockchain = new_blockchain(network, &wallet_opts, &Nodes::None)
+                        .map_err(|e| e.to_string())?;
```
afilini, line 75: You could try moving the blockchain into the `WasmWallet` struct so that you don't have to re-create it every time. It probably needs to be wrapped in a `Rc` like all the other structs.

src/wasm.rs
```diff
@@ -0,0 +1,204 @@
...
+            _ => return Err(Box::<dyn Error>::from("InvalidScriptType")),
+        };
+
+        let descriptor: Result<Descriptor<String>, bdk::Error> = descriptor.translate_pk(
+            |key| Ok(aliases.get(key).unwrap_or(key).into()),
+            |key| Ok(aliases.get(key).unwrap_or(key).into()),
+        );
+        let descriptor = descriptor?;
+
+        Ok(format!("{}", descriptor).into())
```
afilini, line 160: You can also use `descriptor.to_string().into()`

afilini (commented):
I would also add a CI job for wasm (just building the code, no tests) so that we don't forget to keep this updated going forward

notmandatory:
I added this to the 0.7.0 milestone since it's the next big thing worth making a release for.

danielabrozzoni:
The code is almost ready (I'm working on fixing the CI), do you think we could squeeze it in the 0.6.0 release? I'd like to have the playground updated as soon as possible, as if you use it nowadays it generates insecure descriptors :)

.github/workflows/cont_integration.yml
```diff
@@ -55,6 +55,41 @@ jobs:
...
+    name: Build WASM
+    runs-on: ubuntu-latest
+    env:
+      CC: clang-10
+      CFLAGS: -I/usr/include
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Generate cache key
+        run: echo "${{ matrix.rust }}" | tee .cache_key
```
afilini, line 68: These lines are not necessary

src/wasm.rs
```diff
@@ -0,0 +1,200 @@
...
+            Alias::GenExt { extra: path } => {
+                let generated: GeneratedKey<
+                    bitcoin::util::bip32::ExtendedPrivKey,
+                    miniscript::Legacy,
+                > = GeneratableDefaultOptions::generate_default().unwrap();
+
+                let mut xprv = generated.into_key();
+                xprv.network = Network::Testnet;
+
+                format!("{}{}", xprv, path)
```
afilini, line 195: Are you sure there isn't an `m` left behind when printing the derivation path?

src/wasm.rs
```diff
@@ -0,0 +1,200 @@
...
+            "wsh" => Descriptor::new_wsh(policy.compile()?)?,
+            "sh-wsh" => Descriptor::new_sh_wsh(policy.compile()?)?,
+            _ => return Err(Box::<dyn Error>::from("InvalidScriptType")),
+        };
+
+        let descriptor: Result<Descriptor<String>, bdk::Error> = descriptor.translate_pk(
+            |key| Ok(aliases.get(key).unwrap_or(key).into()),
+            |key| Ok(aliases.get(key).unwrap_or(key).into()),
+        );
+        let descriptor = descriptor?;
```
afilini, line 154: You can move the question mark to the end of the previous line

notmandatory:
> The code is almost ready (I'm working on fixing the CI), do you think we could squeeze it in the 0.6.0 release? I'd like to have the playground updated as soon as possible, as if you use it nowadays it generates insecure descriptors :)

Hey sorry I just saw this comment. If this one is just about ready to go then probably better to put it in 0.6.0 than have to do a 0.7.0 right after 0.6.0. That sound OK @rajarshimaitra ?

rajarshimaitra:
Sounds good to me.. If this is almost ready we can push it in 0.6.. I will put in a review too soon..

rajarshimaitra:
On the initial look the code LGTM.. I just have one conceptual question..

Is there any wasm related technical problem of just using this function here and expose it via bindgen? https://github.com/bitcoindevkit/bdk-cli/blob/183e85ec56623090c0efabb09a5173bb452c2d5a/src/handlers.rs#L533

All the wallet structures and everything else is created internally within this function, so in theory we won't need to have dedicated wasm related structure definition. That was my motivation behind simplifying the `handle_command` function to only take a `CliOpts` which is a structured list of strings, same as command line arguments. So a wasm app can just play the function and depending on what it needs to do pass in corresponding string arguments? 

I haven't tried this out myself, so wondering if there's something specific to wasm that doesn't allow us to do that?? The purpose of `handle_command` is one single function to operate everything bdk-cli..

afilini:
For sure the type is not compatible with wasm, you can't get a `CliOpts` directly from wasm.

The second issue is that in wasm only the `MemoryDatabase` is supported, which forces you to use "repl style" commands. `handle_command` recreates the wallet and the database every time, which means you would have an empty database every time and it would be impossible to create transactions, see the balance, etc.

rajarshimaitra:
Thanks @afilini ya that makes sense.. So there is no easy way to do this without the extra structures.. This one is good to go and can be included in the next release itself..

src/wasm.rs
danielabrozzoni: `path` is just a string coming from user input

danielabrozzoni:
> The second issue is that in wasm only the MemoryDatabase is supported, which forces you to use "repl style" commands. handle_command recreates the wallet and the database every time, which means you would have an empty database every time and it would be impossible to create transactions, see the balance, etc.

+1, and also creating the db/blockchain requires creating directories (see `prepare_home_dir`), and you can't do that on Rust

rajarshimaitra (approved):
ACK 731dc74719c8988e5d22358601fed5e10ee4c6ea

PR #129, opened 2022-11-02
Bump bdk version to 0.24.0

notmandatory:
### Description

Update BDK to the latest version, 0.24.0.

### Notes to the reviewers

I also had to add `base64` as a new dependency because it's not exported by `bdk`. 

There's probably a more elegant way to fix the `wasm.rs` compiler function for descriptor.translate_pk, and I don't
know how create a unit test it since it only runs on wasm32. But at least it builds so I think that's close enough for now.

#### All Submissions:


notmandatory:
This is ready to review, but can't be merged until after `bdk` 0.24.0 is released.

notmandatory:
This needs to be updated once bitcoindevkit/bdk-reserves#16 is merged and released.

Cargo.toml
```diff
@@ -12,21 +12,22 @@ readme = "README.md"
...
-bdk = { version = "0.23", default-features = false, features = ["all-keys"]}
+bdk = { version = "0.24", default-features = false, features = ["all-keys"] }
 bdk-macros = "0.6"
 structopt = "0.3"
 serde_json = "1.0"
 log = "0.4"
 zeroize = "<1.4.0"
 dirs-next = "2.0"
 env_logger = "0.7"
+base64 = "^0.13"
```
rajarshimaitra, line 23: Why separately `base64` when `bitcoin` reexports it? I am guessing some version conflicts in the process of upgrade?

But this makes us have two separate base64 dependencies.

```
[[package]]
name = "base64"
version = "0.13.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"

[[package]]
name = "base64-compat"
version = "1.0.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "5a8d4d2746f89841e49230dd26917df1876050f95abafafbe34f47cb534b88d7"
dependencies = [
 "byteorder",
]
```

I don't know in details the diffs between these two crates, but is this intended?

rajarshimaitra (commented):
Sorry took me some time to get back to this.. 

tACK b66773879307a5bebd8578f37a862c99d297262b

Probably I am missing the context of it, but I just have one question..

Cargo.toml
notmandatory: The problem I ran into is that `rust-bitcoin` doesn't seem to re-export the `base64` package, I think now they only re-export crates that they control, so that's why I had to add it.  I copied the `base64` version from the `rust-bitcoin` Cargo.toml:

```toml
[dependencies.base64]
version = "0.13.0"
optional = true
```

rajarshimaitra: Awesome..  It can then be resolved later once the PR is merged.. Opening a ticket to keep us reminded..

rajarshimaitra (approved):
tACK b66773879307a5bebd8578f37a862c99d297262b

PR #128, opened 2022-10-30
Check if psbt is final, before broadcasting

rajarshimaitra:
### Description

Fixes #97

Add a finalizer check before broadcasting psbts. 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #135, opened 2023-01-06
adding hardware signers

ulrichard:
### Description

Adding hardware signers, so that hardware wallets can be used to sign transactions.

### Notes to the reviewers

~~It doesn't work 100% yet for me. 
I am using the following wallet: 
https://github.com/RCasatta/electrum2descriptors/blob/main/tests/wallets2descriptors.rs#L89
https://github.com/RCasatta/electrum2descriptors/blob/main/tests/wallets/multisig_hw_segwit
I can create and sign a transaction with this version of bdk-cli. It displays the details on the Trezor and everything looks good so far. But in the CLI output of bdk-cli it says  "is_finalized= false" and when I open the supposedly signed tx in Electrum, it aslo says "Status: Unsigned". 
So, something must still be missing.~~

Initially I had some dependency problems, so I deleted Cargo.lock. That solved the dependency problems, but that's the reason for the large diff on that file.

## Changelog notice

Added hardware signers through the use of HWI

#### All Submissions:

#### New Features:

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -50,6 +50,11 @@ use bdk::descriptor::Segwitv0;
 use bdk::descriptor::{Descriptor, Legacy, Miniscript};
 #[cfg(all(feature = "reserves", feature = "electrum"))]
 use bdk::electrum_client::{Client, ElectrumApi};
+#[cfg(feature = "hardware-signer")]
+use bdk::hwi::{
+    interface::HWIClient,
+    types::{HWIChain, HWIDescriptor},
+};
 use bdk::keys::bip39::{Language, Mnemonic, WordCount};
 use bdk::keys::DescriptorKey::Secret;
 use bdk::keys::KeyError::{InvalidNetwork, Message};
@@ -57,6 +62,8 @@ use bdk::keys::{DerivableKey, DescriptorKey, ExtendedKey, GeneratableKey, Genera
 use bdk::miniscript::miniscript;
 #[cfg(feature = "compiler")]
 use bdk::miniscript::policy::Concrete;
+#[cfg(feature = "hardware-signer")]
+use bdk::wallet::signer::SignerError;
 use bdk::SignOptions;
 #[cfg(all(feature = "reserves", feature = "electrum"))]
 use bdk::{bitcoin::Address, blockchain::Capability};
@@ -486,6 +493,26 @@ pub(crate) fn handle_key_subcommand(
             } else {
                 Err(Error::Key(Message("Invalid key variant".to_string())))
             }
+        }
+        #[cfg(feature = "hardware-signer")]
+        KeySubCommand::Hardware {} => {
+            let chain = match network {
+                Network::Bitcoin => HWIChain::Main,
+                Network::Testnet => HWIChain::Test,
+                Network::Regtest => HWIChain::Regtest,
+                Network::Signet => HWIChain::Signet,
+            };
+            let devices = HWIClient::enumerate().map_err(|e| SignerError::from(e))?;
+            let descriptors = devices.iter().map(|device_| {
+                let device = device_.clone()
+                    .map_err(|e| SignerError::from(e))?;
+                let client = HWIClient::get_client(&device, true, chain.clone())
+                    .map_err(|e| SignerError::from(e))?;
+                let descriptors: HWIDescriptor<String> = client.get_descriptors(None)
+                    .map_err(|e| SignerError::from(e))?;
+                Ok(json!({"device": device.model, "receiving": descriptors.receive[0].to_string(), "change": descriptors.internal[0]})) 
+            }).collect::<Result<Vec<_>, Error>>()?;
+            Ok(json!(descriptors))
         }
     }
 }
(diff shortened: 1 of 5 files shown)


rajarshimaitra:
Concept ACK.. Thanks for adding this.. I was just thinking about this the other day.. Could be useful if anyone wanna use bdk-cli with mainnet (still not recommended).. 

> I can create and sign a transaction with this version of bdk-cli. It displays the details on the Trezor and everything looks good so far. But in the CLI output of bdk-cli it says "is_finalized= false" and when I open the supposedly signed tx in Electrum, it aslo says "Status: Unsigned".

From an initial glance, it looks like all the wallets you are using are watch-only? It has Xpubs and addresses only.. Such wallets in bdk can't sign for transactions as they don't have the private key.. and `is_finalised=false` means the resulting psbt doesn't have signatures.

PS: We should throw an error in those cases if use asks to `sign` a bdk-cli with watch only descriptors. 

Can you try testing with a single sig wallet, with keys generated from bdk-cli itself?? If the HWI part is working properly that should work.. 

I will also try to test this out in an emulator..

rajarshimaitra:
I think the elctrum error you are seeing is also part of our lack of error hgandling in unsigned psbts and should be fixed with https://github.com/bitcoindevkit/bdk-cli/pull/128. 

But it only handles it while broadcast.. I will try to see how the watch-only case can be captured in signing also..

ulrichard:
I tested with single sig hw wallets now. 
With the Trezor everything works nicely. 
With the Ledger, I got the descriptors, but when I tried to sign, I got an IO error "failed to open device". Anyway, since it works with the Trezor, I'm sure it has nothing to do with the bdk-cli integration.

src/utils.rs
```diff
@@ -496,5 +506,43 @@ where
...
+    let chain = match network {
+        Network::Bitcoin => HWIChain::Main,
+        Network::Testnet => HWIChain::Test,
+        Network::Regtest => HWIChain::Regtest,
+        Network::Signet => HWIChain::Signet,
+    };
+    let devices = HWIClient::enumerate().map_err(|e| SignerError::from(e))?;
+    for device in devices {
+        let _client = HWIClient::get_client(&device, true, chain.clone())
+            .map_err(|e| SignerError::from(e))?;
```
rajarshimaitra, line 533: It seems this `_client` is redundant. Do we need this to operate the signer?

src/utils.rs
```diff
@@ -496,5 +506,43 @@ where
...
+}
+
+/// Add hardware wallets as signers to the wallet
+#[cfg(feature = "hardware-signer")]
+fn add_hardware_signers<D>(wallet: Wallet<D>, network: Network) -> Result<Wallet<D>, Error>
+where
+    D: BatchDatabase,
+{
+    let mut wallet = wallet;
+    //    let chain = HWIChain::from(network);
```
rajarshimaitra, line 523: This doesn't work so can be removed..

rajarshimaitra (changes requested):
Code Review ACK bd2b66ed62e3619a4cd421133bde4bfb2ab22b95

Few more changes and it would be good to go..

Also seems to have conflict with master.. 

I have successfully tried it out with a single sig coldcard..

src/handlers.rs
```diff
@@ -455,6 +459,24 @@ pub(crate) fn handle_key_subcommand(
...
+        KeySubCommand::Hardware {} => {
+            let chain = match network {
+                Network::Bitcoin => HWIChain::Main,
+                Network::Testnet => HWIChain::Test,
+                Network::Regtest => HWIChain::Regtest,
+                Network::Signet => HWIChain::Signet,
+            };
+            let devices = HWIClient::enumerate().map_err(|e| SignerError::from(e))?;
+            let descriptors = devices.iter().map(|device| {
+                let client = HWIClient::get_client(&device, true, chain.clone())
```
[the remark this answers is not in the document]
ulrichard: If multiple hardware wallets are plugged in, the closure is executed multiple times.

src/utils.rs
ulrichard: I still think this line should work, and I could remove the match statement instead. But I couldn't figure out how to make it work.

rajarshimaitra: Ideally, it should. But `HWIChain` doesn't implement `From<Network>`. Thats why its not working.. Probably worth a PR into HWI lib..

src/handlers.rs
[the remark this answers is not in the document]
rajarshimaitra: Ah I see makes sense..

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added hardware signers through the use of HWI
+
```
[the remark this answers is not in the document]
rajarshimaitra: Just to clarify after adding the log in the description, its actually not needed anymore in this file..

rajarshimaitra (commented):
Re Ack 927b06674aeda04bdba6cbad43e09470578ea9ba modulo one comment..

notmandatory:
@ulrichard  we made some CI changes and bumped the bdk and bdk-reserves versions, please rebase and then this should be ready to go. Thanks!

notmandatory (approved):
utACK 100ea8d7e8ae789e3c328706591ec20ba427496b

rajarshimaitra (approved):
ACK 100ea8d7e8ae789e3c328706591ec20ba427496b

PR #140, opened 2023-01-18
bump stable rustc version to 1.65

rajarshimaitra:
### Description

As BDK bumped its stable to 1.65, it is safe for us to move it up also.  

### Notes to the reviewers

`1.65` cause clippy warning that `Eq` isn't derived while `PartialEq` is derived for the commands. We can't derive `Eq` for commands, as there is a `fee_rate` which is `f32` which isn't `Eq`, and this effect cascades to all structures. We also need `PartialEq` for the tests.

A few options were:
 - allow the clippy warning conditionally for `1.65`, `1.57` doesn't have that lint. The way to do that is by creating custom features for different versions and making a build script that activates specific features for specific versions. Feature gate the clippy lint. or use https://github.com/dtolnay/rustversion which is an extra dep.
 
  - Remove `PartialEq` and create a method on `CliOpts` to translate it into `Vec<String>`, and then perform the tests. 
  
  - Make `fee_rate` `u32` and derive `Eq` for all.
  
I went for the last option as it seemed less complex, and not too bad(?). For bdk_cli as majorly a testing tool, it might not need floating point granularity and users can still test out with `u32` fee_rates. Internally it converts it back to `f32`. 

Suggestions welcome.   
  

## Changelog notice

 - Bumped rustc `stable` to 1.65.

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.60.0 # STABLE
+          - 1.65.0 # STABLE
           - 1.57.0 # MSRV
         features:
           - default
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -144,7 +144,7 @@ pub enum CliSubCommand {
 }
 
 /// Backend Node operation subcommands.
-#[derive(Debug, Subcommand, Clone, PartialEq)]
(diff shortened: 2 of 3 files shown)


notmandatory:
I removed the derive for `Eq` on a `CliOpts` and a few others so that we can leave the FeeRate amount as a f32. Clippy for 1.65 is still happy with all the changes.

PR #151, opened 2023-04-03
Added validate_domain Electrum option

### LINKED ISSUE #134: Add `validate_domain` Electrum option
Since https://github.com/bitcoindevkit/bdk/commit/8d4cc3920ab4177fd0f3916155671b1067285eda bdk opens the validate_domain option in electrum config. We should also expose it via bdk-cli electrum options.

anmode:
### Description
fixes #134

#### All Submissions:

#### New Features:

#### Bugfixes:

cc @rajarshimaitra @danielabrozzoni

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -673,6 +673,7 @@ mod test {
                         timeout: None,
                         server: "ssl://electrum.blockstream.info:60002".to_string(),
                         stop_gap: 10,
+                        validate_domain: true,
                     },
                     #[cfg(feature = "esplora")]
                     esplora_opts: EsploraOpts {
@@ -717,6 +718,7 @@ mod test {
                             "--change_descriptor", "wpkh(tpubDEnoLuPdBep9bzw5LoGYpsxUQYheRQ9gcgrJhJEcdKFB9cWQRyYmkCyRoTqeD4tJYiVVgt6A3rN6rWn9RYhR9sBsGxji29LYWHuKKbdb1ev/1/*)",
                             "--server","ssl://electrum.blockstream.info:50002",
                             "--stop_gap", "20",
+                            "--validate_domain", "true",
                             "get_new_address"];
 
         let cli_opts = CliOpts::from_iter(&cli_args);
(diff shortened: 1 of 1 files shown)


anmode:
Sorry! Now after testing I pushed. I  Hope all check to pass.

notmandatory:
Hi, thanks for adding a `validate_domain` option, can you make it a flag instead of something that takes a true or false value? See here: https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#flags

anmode:
Ohh okay!! I tried!. Do not know if that works or not.

src/commands.rs
```diff
@@ -320,6 +320,10 @@ pub struct ElectrumOpts {
         default_value = "10"
     )]
     pub stop_gap: usize,
+
+    /// Enable or disable domain validation when connecting to Electrum servers.
+    #[clap(name = "VALIDATE_DOMAIN", long = "validate_domain", takes_value(false))]
```
notmandatory, line 325: you shouldn't need `takes_value(false)`
```suggestion
    #[clap(name = "VALIDATE_DOMAIN", long = "validate_domain")]
```

src/commands.rs
```diff
@@ -320,6 +320,10 @@ pub struct ElectrumOpts {
...
+    pub validate_domain: Option<bool>,
```
notmandatory, line 326: This shouldn't be an `Option`, only a `bool`, if the flag isn't used then the value will default to `false`.
```suggestion
    pub validate_domain: bool,
```

notmandatory (changes requested):
You should always run the tests locally to verify your code changes are correct.

src/commands.rs
anmode: oh, Okayy

src/commands.rs
anmode: sure!

anmode:
Okayy. Actually i am new so didn't knew about what testing command are. As i did `cargo fmt` , `cargo clippy` and `cargo test`. All goes well.

notmandatory:
Very old and needs rebase. Closing for now and can reopen later if author wants to.

PR #155, opened 2023-06-09
Added validate_domain Electrum option

dspicher:


notmandatory:
Take a look at the utils.rs file and on line 403 you need to use this new param you're adding.

src/commands.rs
```diff
@@ -320,6 +320,10 @@ pub struct ElectrumOpts {
         default_value = "10"
     )]
     pub stop_gap: usize,
+
+    /// Enable domain validation when connecting to Electrum servers.
+    #[clap(name = "VALIDATE_DOMAIN", long = "validate_domain")]
```
notmandatory, line 325: In the `utils.rs` the default for this value is true, as you have it here the user will be required to set the value to `true` or `false`. It's better if you set the clap `default_value` to true then the user only needs to do something if they want it to be false.

dspicher:
> Take a look at the utils.rs file and on line 403 you need to use this new param you're adding.

Ah yeah, makes sense. Fixed.

src/commands.rs
dspicher: Done. Also added a `v` short option.

notmandatory (approved):
ACK 976373d257b52ba8fba8e5f24088d666816a9947

src/commands.rs
```diff
@@ -320,6 +320,14 @@ pub struct ElectrumOpts {
         default_value = "10"
     )]
     pub stop_gap: usize,
+
+    /// Enable domain validation when connecting to Electrum servers.
```
notmandatory, line 324: Need to make it an Option<bool> and set the default in the code, see below.

```suggestion
    /// Enable domain validation when connecting to Electrum servers  [default: true]
```

src/commands.rs
```diff
@@ -320,6 +320,14 @@ pub struct ElectrumOpts {
...
+    #[clap(
+        name = "VALIDATE_DOMAIN",
+        long = "validate_domain",
+        default_value = "true"
```
notmandatory, line 328: Remove default.
```suggestion
```

src/commands.rs
```diff
@@ -320,6 +320,14 @@ pub struct ElectrumOpts {
...
+    )]
+    pub validate_domain: bool,
```
notmandatory, line 330: Change type to Option:
```suggestion
    pub validate_domain: Option<bool>,
```

src/utils.rs
```diff
@@ -400,7 +400,7 @@ pub(crate) fn new_blockchain(
             retry: wallet_opts.proxy_opts.retries,
             timeout: wallet_opts.electrum_opts.timeout,
             stop_gap: wallet_opts.electrum_opts.stop_gap,
-            validate_domain: true,
+            validate_domain: wallet_opts.electrum_opts.validate_domain,
```
notmandatory, line 403: Default to true here:
```suggestion
            validate_domain: wallet_opts.electrum_opts.validate_domain.unwrap_or(true),
```

You'll also need to fix the corresponding tests.

notmandatory (changes requested):
I tested this and realized the approach won't work since clap treats `bool` as a "flag" with no way to set to `false` if the default is `true`. See comments for how to fix.

src/commands.rs
dspicher: Done.

src/commands.rs
dspicher: Done.

src/commands.rs
dspicher: Done.

src/utils.rs
dspicher: Done.

notmandatory:
This is old and needs to be rebased. Happy to re-open if author would like to update it.

Issue #172, opened 2025-01-10
Update to use `bdk_wallet` 1.0.0


notmandatory:
**Describe the enhancement**  

This project is overdue for an update to use the latest and greatest `bdk_wallet` 1.0.0. This will be the tracking issue for this update plus some other refactoring and feature enhancements needed for this app.

- [x] Update to `bdk_wallet`, remove sled support, change rust version to `stable`
- [x] Support chain data clients: electrum, esplora, rpc, kyoto (CBF) in same binary
- [x] Add page in book-of-bdk for how to install and use `bdk-cli`

**Use case**  

I'd like this project to be a tool for manually testing `bdk_wallet` and training new devs how the `bdk_wallet` APIs work in a simple CLI app.

**Additional Context**

The `bdk` repo already contains a simple cli using the `bdk_chain` crate, but it is only meant to be an example and not a supported app.

tvpeter:
@notmandatory I would like to work on this

notmandatory:
Great! I've put your name on it @tvpeter .

PR #181, opened 2025-05-03
feat(cbf): add `cbf` feature using `bdk_kyoto`

tvpeter:
### Description

 This PR re-enables the Compact Block Filters (cbf) feature using `bdk_kyoto` and it is part of updating the library
to use the latest `bdk` crates.

### Notes to the reviewers

This is part of issue  #172 

### Checklists


notmandatory:
Good job! I did a quick review and looks good to me. I still need to do some actual testing and see how it works from the cli. 

@rustaceanrob if you have time please let us know if this is the reasonable approach to adding cbf to the bdk-cli test app.

src/commands.rs
```diff
@@ -247,7 +254,7 @@ pub struct CompactFilterOpts {
         long = "cbf-skip-blocks",
         default_value = "0"
     )]
-    pub skip_blocks: usize,
+    pub skip_blocks: u32,
```
rustaceanrob, line 257: I would recommend making this an optional, as the node builder will normally use the `Wallet` height, unless configured otherwise. This will be made clearer in the next `bdk_kyoto` release

src/handlers.rs
```diff
@@ -431,6 +418,105 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                    }
+                    tracing::info!("Kyoto node is running");
+
+                    tracing::info!(
+                        "Initial derivation indices: External={}, Internal={}",
+                        wallet.derivation_index(KeychainKind::External).unwrap_or(0),
+                        wallet.derivation_index(KeychainKind::Internal).unwrap_or(0)
+                    );
+
+                    tracing::info!("Adding wallet scripts to node");
```
rustaceanrob, line 452: This part is handled by the builder and may be safely removed

src/utils.rs
```diff
@@ -178,6 +191,21 @@ pub(crate) fn new_blockchain_client(wallet_opts: &WalletOpts) -> Result<Blockcha
                 client: Box::new(client),
             }
         }
+
+        #[cfg(feature = "cbf")]
+        ClientType::Cbf => {
+            // default to skipping 10 blocks away from the wallet chain tip
+            let skip_blocks = if wallet_opts.compactfilter_opts.skip_blocks == 0 {
```
rustaceanrob, line 198: This is handled by the node builder. In relation to the optional scan height, we should only check that the user is trying to recover from `Some` birthday height. Otherwise you can just use the builder directly and it will account for the wallet height

src/handlers.rs
```diff
@@ -537,6 +681,113 @@ pub(crate) async fn handle_online_wallet_subcommand(
                 RpcClient { client } => client
                     .send_raw_transaction(&tx)
                     .map_err(|e| Error::Generic(e.to_string()))?,
+
+                #[cfg(feature = "cbf")]
+                KyotoClient { client } => {
```
rustaceanrob, line 686: If it is alright with @notmandatory, I would prefer leaving this `unimplemented!` for now, and I will update this logic with new transaction logic coming at the end of the week. It will greatly simplify things here

tvpeter:
Thank you for the review @rustaceanrob. I will update and revert.

tvpeter:
> Gave it one more read. There will be a fair amount of code churn to update to the new `bdk_kyoto` if this PR is merged. Would you be able to update this PR to `bdk_kyoto = 0.9.0`? There wouldn't be any follow-up required if we start on the latest version. Feel free to message me on discord if you need help

Yes, I've just seen that you've released version 0.9.0. I will update to 0.9.0 to avoid revisiting this issue.I will message you if I need any clarification. Thank you.

src/utils.rs
```diff
@@ -263,3 +295,96 @@ pub(crate) fn new_wallet(network: Network, wallet_opts: &WalletOpts) -> Result<W
...
+    if !requester.is_running() {
+        tracing::error!("Kyoto node is not running");
+        return Err(Error::Generic("Kyoto node failed to start".to_string()));
+    }
+    tracing::info!("Kyoto node is running");
+
+    let mut updates_applied = false;
+    let mut initial_tx_count = wallet.transactions().count();
+
+    loop {
```
rustaceanrob, line 356: I think we will want to just call `update_subscriber.update()` a single time without a loop. This will reflect the other blockchain clients closer, where there is a single sync and update returned. This should also simplify the rest of the syncing logic here.

rustaceanrob:
Nice job. One more review from me but otherwise looks good. I think you can also squash these to a single commit with `git rebase -i HEAD~3`. They all seem related to each other

src/utils.rs
tvpeter: I initially thought that syncing required fetching multiple updates, but that doesn't seem to be the case. However, I noticed that it sometimes randomly hangs without providing further information. I checked for an option to configure a timeout for the syncing operation, but I couldn't find any related configuration option. It seems that users may need to manually terminate the process and restart it.

rustaceanrob: If you attempt to sync from a block height that is lower than the current lowest in the block header chain, the node will fail to proceed. This is to prevent users from moving backward in the chain history. Are there any further details on the conditions of the hang? There were no logs written?

tvpeter: If that's the case, I don't think users should be allowed to set the `skip-blocks` parameter, as they won't know their chain tip.

The following are the only logs displayed by the node.
Screenshot 2025-05-07 at 8 19 42 PM

tvpeter: And the `skip-blocks` parameter was not set, so it used the default `Sync` scan type.

rustaceanrob: > as they won't know their chain tip

If the wallet reports a higher sync height, the recovery will be ignored, as it already occurred. We can discuss more on discord if you would like.

> The following are the only logs displayed by the node. 

The node cannot proceed without any peers to connect to. If your DNS query fails, which may occur on public WiFi or low bandwidth connections, the node has no ability to sync without using configured peers.

tvpeter: I see. Thank you alot.

rustaceanrob:
ACK 6debc68afaf545f497cdd17f0e79172313d40463

Successfully ran a local sync. Thanks man 😎

notmandatory:
What does the "Unexpected invalid proof of work when importing a block header..." message mean here?

```
Running `target/debug/bdk-cli -n testnet4 wallet -d sqlite -c cbf sync`
2025-05-08T17:48:17.774732Z  INFO bdk_cli::utils: Kyoto node is running
2025-05-08T17:48:17.783905Z  INFO bdk_cli::utils: Starting node
2025-05-08T17:48:17.783917Z  INFO bdk_cli::utils: Configured connection requirement: 4 peers
2025-05-08T17:48:17.783922Z  INFO bdk_cli::utils: Attempting to load headers from the database
2025-05-08T17:48:17.783927Z  INFO bdk_cli::utils: Unexpected invalid proof of work when importing a block header. expected 419766864, got: 486604799
2025-05-08T17:48:17.783963Z  INFO bdk_cli::utils: Received update: applying to wallet
2025-05-08T17:48:17.784046Z  INFO bdk_cli::utils: Chain tip: 81476, Transactions: 0, Balance: 0
2025-05-08T17:48:17.784053Z  INFO bdk_cli::utils: Sync completed: tx_count=0, balance=0
{}
```

notmandatory:
One nit, can we move the `light_client_data` the the safe folder where the rest of the wallet data is kept? Can be done in a follow-up PR also.

notmandatory (commented):
ACK 6debc68afaf545f497cdd17f0e79172313d40463

Great work @tvpeter and thanks @rustaceanrob for your detailed review.

notmandatory (approved):

tvpeter:
> One nit, can we move the `light_client_data` the the same folder where the rest of the wallet data is kept? Can be done in a follow-up PR also.

Yes, I will do that.

PR #182, opened 2025-05-09
Implement transaction broadcasting for Kyoto

rustaceanrob:
The actual implementation comes down to listening for an info message that reports the transaction was sent to a peer. For simplicity I am ignoring any wallet updates, but if the user calls the `Sync` command they can catch them. Follows up #181 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -48,7 +48,11 @@ use std::str::FromStr;
 
 #[cfg(feature = "electrum")]
 use crate::utils::BlockchainClient::Electrum;
+#[cfg(feature = "cbf")]
+use bdk_kyoto::{Info, LightClient};
 use bdk_wallet::bitcoin::base64::prelude::*;
+#[cfg(feature = "cbf")]
+use tokio::select;
 #[cfg(any(
     feature = "electrum",
     feature = "esplora",
@@ -507,7 +511,6 @@ pub(crate) async fn handle_online_wallet_subcommand(
                 (Some(_), Some(_)) => panic!("Both `psbt` and `tx` options not allowed"),
                 (None, None) => panic!("Missing `psbt` and `tx` option"),
             };
-
             let txid = match client {
                 #[cfg(feature = "electrum")]
                 Electrum {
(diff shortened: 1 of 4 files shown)


src/handlers.rs
```diff
@@ -531,8 +534,49 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                        node,
+                    } = client;
+
+                    let subscriber = tracing_subscriber::FmtSubscriber::new();
+                    tracing::subscriber::set_global_default(subscriber)
+                        .map_err(|e| Error::Generic(format!("SetGlobalDefault error: {}", e)))?;
+
+                    tokio::task::spawn(async move { node.run().await });
+                    tokio::task::spawn(async move {
+                        select! {
```
tvpeter, line 553: Is it possible to use the `trace_logger` fn here by using an MPSC channel to forward Info messages, with two receivers: one for trace_logger and one for the main task?

tvpeter (commented):
Probably my network is poor, I have not been able to successfully broadcast a tx but will keep trying.

Thank you for helping with this.

src/handlers.rs
rustaceanrob: > two receivers

Only one receiver is allowed with `mpsc`. Because the `Info` channel is already passed to us, I think it is best to use it directly

rustaceanrob:
> Probably my network is poor, I have not been able to successfully broadcast a tx but will keep trying.

I am working now to improve the broadcasting reliability in Kyoto, so there is a chance you are running into bugs of mine. However the code here wouldn't change, only internals of the node.

src/handlers.rs
tvpeter: Gotten. Thank you.

tvpeter:
>   only internals of the node.

Would that require updating `bdk-kyoto`?

rustaceanrob:
I believe `bdk_kyoto` would require a patch release as well. I should remark that I have successfully sent transactions with Kyoto over test networks and Bitcoin, so the improvements are more for reliability.

notmandatory:
I also was unable to get it to broadcast (I'm using testnet4). It looks like it's never getting to this line:
```rust
tracing::info!("Successfully gossiped wtxid: {wtxid}");
```

rustaceanrob:
I think I've ironed out what is happening with [this test](https://github.com/2140-dev/kyoto/pull/377/files). Because I don't allow for any buffer of time between the node starting and the attempting to send the transaction, Kyoto advertises the transaction sometime during the version handshake with the remote node. As a result the remote node just drops it on the floor. I will change this to broadcast as the node is downloading filters, at which point it will be thoroughly connected.

tvpeter:
> Because I don't allow for any buffer of time between the node starting and the attempting to send the transaction, Kyoto advertises the transaction sometime during the version handshake with the remote node. As a result the remote node just drops it on the floor. 

That’s exactly what I thought. If there were a way to determine the number of established connections before attempting a broadcast, it would have simplified the process. That is why in the initial implementation, I allowed a 60-second buffer before broadcasting, which made it a bit more reliable.

rustaceanrob:
There is `Info::ConnectionsMet`, but that is only emitted once the total requirement is met, but it would make the broadcast more reliable. I am going to add a short `sleep`, then await the `TxGossiped` with a timeout. If it fails for some reason, the user can try again. Thanks for the observation!

rustaceanrob:
I was able to successfully broadcast a transaction with the most recent commit. I changed the default connections to 2, as there are only a couple CBF nodes on testnet4 and signet

src/handlers.rs
```diff
@@ -423,7 +427,7 @@ pub(crate) async fn handle_online_wallet_subcommand(
             Ok(json!({}))
         }
         Sync => {
-            #[cfg(any(feature = "electrum", feature = "esplora"))]
+            #[cfg(any(feature = "electrum", feature = "esplora", feature = "cbf"))]
```
notmandatory, line 430: This line doesn't need the `cbf` feature gate since `request` isn't used by kyoto and having it causes a unused warning during build if only the `cbf` feature is enabled.
```suggestion
            #[cfg(any(feature = "electrum", feature = "esplora"))]
```

src/handlers.rs
```diff
@@ -531,8 +534,69 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                                break;
+                            }
+                            _ => tracing::info!("{info}"),
+                        }
+                    }
+                    tokio::time::timeout(tokio::time::Duration::from_secs(15), async move {
+                        while let Some(info) = info_subscriber.recv().await {
+                            match info {
+                                Info::TxGossiped(wtxid) => {
+                                    tracing::info!("Succuessfully broadcast WTXID: {wtxid}");
```
notmandatory, line 583: Typo:
```suggestion
                                    tracing::info!("Successfully broadcast WTXID: {wtxid}");
```

notmandatory (commented):
Even with the latest commit I was unable to broadcast a Tx. I did a little debugging and found that I get `info_subscriber.len() = 0` right before the `while let Some(info) = info_subscriber.recv().await { ...` code is called which means it returns `None` and nothing is ever broadcast.

@rustaceanrob what OS are you using? I'm on MacOS (M1). Could it be some timing difference or way tokio is handling multiple threads? I also tried adding a sleep prior to the above line and it didn't make a difference.

rustaceanrob:
It sounds like your node is returning an error when calling "run". I suggest removing the light_client_data directory. The only time I've gotten errors is 1. the light_client_data directory is in use by another program 2. the wallet is requesting a sync lower than the headers Kyoto is able to load from SQL

notmandatory:
You were right! I cleared the light_client_data directory and re-sync'd skipping the first 50K blocks and then I was able to broadcast. 

I probably ran into this because the wallet data and light_client_data were in different spots and I probably removed one and not the other during testing, #183 puts them under the same parent.  I still think this could use some better error reports to see when issues like this come up, but that can come in future PRs.

notmandatory (approved):
ACK 8ac6957ffea79f42c243bb02d5fc789b923f18bc

Please just rebase and fix the build warning and typo and then I'll merge it. Or if you're busy let me know and I'll do it. Thanks!

src/handlers.rs
notmandatory: fixed.

src/handlers.rs
notmandatory: Fixed.

PR #197, opened 2025-05-29
Update bdk_wallet dependency to 2.0.0 plus related

notmandatory:
### Description

* Updated dependencies: 
  * bdk_wallet to 2.0.0
  * bdk_bitcoind_rpc to 0.20.0
  * bdk_electrum to 0.23.0
  * bdk_esplora to 0.22.0
  * bdk_kyoto to 0.11.0

* Renamed `BuilderError` to `KyotoBuilderError`
* Added `KyotoUpdateError`

### Notes to the reviewers

There are still some unrelated warnings in the test code that should be fixed in a followup PR.

## Changelog notice

* Updated project to use `bdk_wallet` 2.0 and related supporting chain client versions.

#### All Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,12 +79,12 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.7"
+version = "3.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
+checksum = "6680de5231bd6ee4c6191b8a1325daa282b415391ec9d3a37bd34f2060dc73fa"
 dependencies = [
  "anstyle",
- "once_cell",
+ "once_cell_polyfill",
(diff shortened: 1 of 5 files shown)


notmandatory:
@rustaceanrob let me know when you get an updated `bdk_kyoto` and I'll test it out here too.

rustaceanrob:
Looks like there is a bug for Testnet4 that does not allow for the 20 minute exception. If you switch to Signet I expect everything to work properly. I'll be able to work on a patch in a week or two

notmandatory:
Makes sense, thanks for taking a look and good to know it's just a `testnet4` thing you can fixup later.

tvpeter:
tACK bd526bdf91be3c44704b76ea43e29ef01f6b81a6

> I confirmed CBF/Kyoto sync and receive test works fine on `signet`.

I've tested `signet` and `testnet4` on `electrum` and they worked well. Kyoto do get stuck for me establishing connections but Rob previously explained that it might be due to poor connectivity. So since yours worked, I'm confident it'll work for me too.

PR #200, opened 2025-06-04
feat: add Payjoin v2 send online command with fallback to v1

mehmetefeumit:
### Description

This change implements the online command for **sending** a Payjoin. In the context of Payjoin, sending refers to creating the original PSBT which can be used as the vanilla transaction, sending it to the receiver for their input's addition, and broadcasting the Payjoin proposal which the receiver sends back.

The communication between the sender and the receiver can be done synchronously (v1) or asynchronously (v2) through a OHTTP relay as the middle man. This implementation supports both: if the OHTTP relay is not defined or there is an issue with the v2 process, it will use/fallback to v1. Otherwise, it will use v2.

#### Persistence
Note that this change does not implement persistence. If a sending session is started and is left during the process (i.e. when the sender is waiting for the receiver to contribute their UTXOs and send back the Payjoin proposal), then the user cannot pick up from where they left of.

I made the decision myself due to the UX complexity of adding the `resume` functionality. bdk-cli only has two layers of commands. For example, you can implement `bitcoin-cli wallet send_payjoin`, but you cannot do `bitcoin-cli wallet send_payjoin resume`. So there is a UX/interface decision to be made on how the resume would be implemented.

We could do `send_payjoin`, `retrieve_payjoin`, and `resume_send_payjoin`, etc., but that is also a hit on the complexity of the interface and the UX of Payjoin in general. Same with adding another flag for resuming, but the more I thought about it, the more complex it got...

So I left it as Noop persistence for now, but do let me know if you have ideas!

### Notes to the reviewers

Tested on my local regtest with rpc, both v1 and v2, using the OHTTP relays and directories in the [payjoin-cli README](https://github.com/payjoin/rust-payjoin/tree/master/payjoin-cli). Feel free to do the same, and let me know if there are comments, documentation, etc. which should be changed with this. I was not able to find any...

Here's a sample command you can use to test on regtest:
```
cargo run --features=rpc -- -n regtest wallet -a USERNAME:PASSWORD -w WALLET_NAME -e DESCRIPTOR send_payjoin --ohttp_relay RELAY --fee_rate 1 --uri URI
```

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk-cli/blob/master/CONTRIBUTING.md)
* [X] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [] I've added tests for the new feature  <--- There is no testing suite for this.
* [X] I've added docs for the new feature
* [N/A] I've updated `CHANGELOG.md`

#### Bugfixes:


src/handlers.rs
```diff
@@ -511,94 +517,207 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
-                } => client
-                    .broadcast(&tx)
-                    .await
-                    .map(|()| tx.compute_txid())
-                    .map_err(|e| Error::Generic(e.to_string()))?,
-                #[cfg(feature = "rpc")]
-                RpcClient { client } => client
-                    .send_raw_transaction(&tx)
-                    .map_err(|e| Error::Generic(e.to_string()))?,
+            let txid = broadcast_transaction(client, tx).await?;
```
mehmetefeumit, line 714: Moved the client matching and broadcasting logic outside the command code since it is now shared between the `Broadcast` and `SendPayjoin` commands.

spacebear21: It will help reviewers to break up this PR into individual atomic commits; for example this move should be a separate commit that simply "extracts shared broadcast transaction logic" so that it can be easily checked that it does what it says. See also this great resource: https://cbea.ms/git-commit/

src/commands.rs
```diff
@@ -414,6 +414,18 @@ pub enum OnlineWalletSubCommand {
...
+    SendPayjoin {
+        /// BIP 21 URI to send to.
+        #[arg(env = "URI", long = "uri", required = true)]
+        uri: String,
+        /// Fee rate to use in sat/vbyte.
+        #[arg(env = "SATS_VBYTE", short = 'f', long = "fee_rate", required = true)]
+        fee_rate: u64,
+        /// URL of the OHTTP relay.
+        #[arg(env = "PAYJOIN_OHTTP_RELAY", long = "ohttp_relay")]
+        ohttp_relay: Option<String>,
```
spacebear21: `payjoin-cli` uses a [list of ohttp relays](https://github.com/payjoin/rust-payjoin/blob/master/payjoin-cli/src/app/config.rs#L36) so that it can fallback to another relay in case one is not working, it may be useful to follow the same pattern for bdk-cli instead of an Option.

src/utils.rs
```diff
@@ -369,3 +369,17 @@ pub async fn sync_kyoto_client(wallet: &mut Wallet, client: LightClient) -> Resu
 
     Ok(())
 }
+
+#[cfg(any(
+    feature = "electrum",
+    feature = "esplora",
+    feature = "cbf",
+    feature = "rpc"
+))]
```
spacebear21: This feature gate seems unnecessary for this function

src/handlers.rs
```diff
@@ -511,94 +517,207 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+            for (proposal_txin, proposal_psbt_input) in input_pairs(&mut psbt) {
+                if let Some((orig_txin, orig_psbt_input)) = original_inputs_iter.peek() {
+                    if proposal_txin.previous_output == orig_txin.previous_output {
+                        proposal_psbt_input.witness_utxo = orig_psbt_input.witness_utxo.clone();
+                        proposal_psbt_input.non_witness_utxo =
+                            orig_psbt_input.non_witness_utxo.clone();
+                        original_inputs_iter.next();
+                    }
+                }
+            }
```
spacebear21: I don't think it's true that the receiver strips the utxo information? See [here](https://github.com/payjoin/rust-payjoin/blob/ac66657582edd0efe38e6c4f95f65e3588a608e3/payjoin/src/receive/v1/mod.rs#L699).

The only thing that gets stripped are the [sender signatures](https://github.com/payjoin/rust-payjoin/blob/ac66657582edd0efe38e6c4f95f65e3588a608e3/payjoin/src/receive/v1/mod.rs#L757) afaict.

DanGould: I'm not sure that a receiver *necessarily* strips it but I'm not sure you can depend on them not being stripped. Since BDK needs the info to find the keys associated with a script, you might want to add it back in as is done here.

Cargo.toml
```diff
@@ -21,6 +21,8 @@ log = "0.4"
 serde_json = "1.0"
 thiserror = "2.0.11"
 tokio = { version = "1", features = ["full"] }
+payjoin = { version = "0.23.0", features = ["v1", "v2", "io"] }
+minreq = { version = "2.13.2", features = ["https"] }
```
DanGould: If you're using the `io` feature, you're already depending on `reqwest`, so minreq is a second HTTP client. If you want to use minreq to reduce the dependency burden I recommend also writing the `payjoin/io` feature functions using `minreq`.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                .amount
+                .ok_or_else(|| Error::Generic("Amount is not specified in the URI.".to_string()))?;
+            let fee_rate =
+                FeeRate::from_sat_per_vb(fee_rate).expect("Provided fee rate is not valid.");
+
+            // This original PSBT will be sent to the receiver for their UTXO contributions.
+            let mut original_psbt = {
+                let mut tx_builder = wallet.build_tx();
+                tx_builder
+                    .ordering(bdk_wallet::TxOrdering::Untouched)
```
DanGould: using this non-default, non-shuffled TxOrdering creates a wallet fingerprint. I do not recommend doing so.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                    ))
+                })?
+            };
 
+            if !wallet.sign(
+                &mut original_psbt,
+                SignOptions {
+                    try_finalize: true,
+                    ..Default::default()
+                },
```
DanGould:
```suggestion
               SignOptions::default()
```

`try_finalize: true` is default behavior.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                ))
+            })?;
+            let req_ctx = Sender::load(storage_token, &persister)
+                .map_err(|e| Error::Generic(format!("Failed to load the request context using the storage token and the persister {}", e)))?;
+
+            let mut psbt = {
+                // Payjoin URI will contain a hashtag fragment separator for Payjoin v2
+                if uri.extras.endpoint().to_string().contains('#') {
+                    println!("Attempting a Payjoin v2 send...");
+                    match ohttp_relay {
```
DanGould: `req_ctx.extract_v2(relay_url)` will error if v2 is not supported. I would use that behavior instead of checking for a `#`.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                                println!("Trying OHTTP relay: {}", relay);
+
+                                // Using a closure here to stop the error from propagating.
+                                // The check of the attempt result will be done later and swallowed
+                                // so that we can attempt the send with the next relay in the list.
+                                let attempt_result = (|| -> Result<_, Error> {
+                                    let relay_url = Url::parse(relay.as_str()).map_err(|e| {
+                                        Error::Generic(format!("Failed to parse relay URL: {}", e))
+                                    })?;
+
```
DanGould: I would probably parse the error arguments *before* creating an original_psbt, since the whole function will fail otherwise. IMO the ohttp_relay logic should be isolated from this psbt block.

src/utils.rs
```diff
@@ -369,3 +369,17 @@ pub async fn sync_kyoto_client(wallet: &mut Wallet, client: LightClient) -> Resu
...
+    feature = "esplora",
+    feature = "rpc",
+    feature = "cbf",
+))]
+pub fn send_payjoin_post_request(req: payjoin::Request) -> Result<minreq::Response, Error> {
+    minreq::post(req.url)
+        .with_header("Content-Type", req.content_type)
+        .with_body(req.body)
+        .send()
+        .map_err(|e| Error::Generic(format!("HTTP request for Payjoin failed: {}", e)))
```
DanGould: You can return a `minreq` error here (or reqwest if you replace it`. The generic error constraint is at the application level, not internal here.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                                    let v2_ctx =
+                                        ctx.process_response(response.as_bytes()).map_err(|e| {
+                                            Error::Generic(format!(
+                                                "Failed to initialize v2 context: {}",
+                                                e
+                                            ))
+                                        })?;
+
+                                    let v2_ctx = Arc::new(v2_ctx);
```
DanGould: Why wrap in `Arc`?

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+                                        result = Some(psbt);
+                                        break;
+                                    }
+                                    Err(e) => {
+                                        println!("Payjoin v2 failed with relay '{}': {}", relay, e);
+                                        // Swallow the error so that we continue with the next
+                                        // relay. We will eventually fail later if result ends up
+                                        // being an Err.
+                                    }
                                 }
```
DanGould: It seems that there are lots of errors that are unrelated to relays that might propagate here. I recommend abstracting relay management behavior from other state machinery so that such a handler only gets triggered if there was an issue with a relay.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
                             }
+
+                            // If the result is still Err, then all OHTTP relays failed.
+                            result.ok_or_else(|| {
+                                Error::Generic("Exhausted all OHTTP relays for Payjoin v2 send. Failing the operation.".to_string())
+                            })?
                         }
+                        _ => {
+                            return Err(Error::Generic("Payjoin v2 URI detected but no OHTTP relays provided. Payjoin v2 requires at least one OHTTP relay URL to be provided.".to_string()));
                         }
```
DanGould: As stated above validating the arguments before proceeding into logic will make for more readable code and a more sensible user experience.

src/handlers.rs
```diff
@@ -524,16 +530,234 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
                     }
+                } else {
+                    println!("Attempting a Payjoin v1 send...");
```
DanGould: I don't think it makes sense to attempt a payjoin v1 send fails for any reason. .extract_v2 exposes variants of failure from an invalid v2 address. I do believe the parsing needs to be improved there so it's abundantly clear if the URI is v1/v2 capable, but I do know that we shouldn't proceed to v1 if any failure happens in the v2 process, only if the v2 URI is invalid (as evidenced by `CreateRequestError` from `extract_v2`) but a v1 request may still be extracted.

Cargo.toml
```diff
@@ -21,6 +21,8 @@ log = "0.4"
 serde_json = "1.0"
 thiserror = "2.0.11"
 tokio = { version = "1", features = ["full"] }
+payjoin = { version = "0.23.0", features = ["v1", "v2", "io"] }
```
DanGould: I don't believe you need the `v1` feature, you want to use the backwards compatible `v2` feature to handle `v1` rather than the exclusive `v1` feature.

src/handlers.rs
```diff
@@ -523,85 +527,199 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+            // We will go with Payjoin v2 if there is at least one functional OHTTP relay passed as
+            // an argument and the request-context extraction is successful. Otherwise, if there is
+            // an error in finding a healthy relay or extracting the v2 context, we use v1.
+            let v2_setup = if ohttp_relays.is_empty() {
+                Err("No OHTTP relays provided.".to_string())
+            } else {
+                async {
+                    for relay in ohttp_relays {
+                        if let Ok(_) =
+                            payjoin::io::fetch_ohttp_keys(relay.clone(), req_ctx.endpoint()).await
```
mehmetefeumit: This is similar to what payjoin-cli is doing [here](https://github.com/payjoin/rust-payjoin/blob/master/payjoin-cli/src/app/v2/mod.rs#L257). The main difference is that we are not really using the keys here since we are not storing them anywhere. Currently, the entire implementation is stateless -- there is no resume functionality, and we do not store information on the OHTTP relays either. So we're just using the `fetch...` function here to determine if a OHTTP relay passed in the arguments is healthy. And then scrap the keys, and leave everything to the `extract_v2` call.

src/payjoin/mod.rs
```diff
@@ -45,6 +45,648 @@ impl<'a> PayjoinManager<'a> {
...
+                    .map_err(|e| {
+                        ImplementationError::from(
+                            format!("Error occurred when signing the Payjoin PSBT: {}", e).as_str(),
+                        )
+                    })?
+                {
+                    return Err(ImplementationError::from(
+                        "Failed to sign and finalize the Payjoin proposal PSBT.",
+                    ));
+                }
```
Mshehu5: Seems this always returns false because sender's inputs can't be signed so it keeps getting error: Failed to sign and finalize the Payjoin proposal PSBT.

Mshehu5 (commented):
Awesome work Efe!

While testing this, I kept running into the error: "Failed to sign and finalize the Payjoin proposal PSBT." on the receiver side from what I observed, the issue comes from the wallet.sign() call inside the finalize_proposal function. The if statement expects a true value but it always evaluates to false because it can’t sign the sender’s input in the PSBT. This results in an error.

Once I removed that check, the Payjoin flow worked. We might want to consider a better or more appropriate way to perform this validation.

Feel free to test or confirm this I might be wrong and it could be something on my end.

src/payjoin/mod.rs
mehmetefeumit: Thanks a lot for the catch! It is fixed.

Cargo.toml
```diff
@@ -29,8 +31,9 @@ bdk_esplora = { version = "0.22.1", features = ["async-https", "tokio"], optiona
 bdk_kyoto = { version = "0.15.1", optional = true }
 bdk_redb = { version = "0.1.0", optional = true }
 shlex = {  version = "1.3.0", optional = true }
-tracing = "0.1.41"
-tracing-subscriber = "0.3.20"
+payjoin = { version = "1.0.0-rc.0", features = ["v1", "v2", "io", "_test-utils"], optional = true}
```
spacebear21: `1.0.0-rc.1` is the latest RC

Cargo.toml
```diff
@@ -43,10 +46,13 @@ sqlite = ["bdk_wallet/rusqlite"]
...
-electrum = ["bdk_electrum"]
-esplora = ["bdk_esplora"]
-rpc = ["bdk_bitcoind_rpc"]
+cbf = ["bdk_kyoto", "_payjoin-dependencies"]
+electrum = ["bdk_electrum", "_payjoin-dependencies"]
+esplora = ["bdk_esplora", "_payjoin-dependencies"]
+rpc = ["bdk_bitcoind_rpc", "_payjoin-dependencies"] 
+
+# Internal features
+_payjoin-dependencies = ["payjoin", "reqwest", "url"]
```
spacebear21, line 55: I believe the `url` dependency was removed from the public payjoin API for 1.0.0-rc, so it should no longer be required.

src/commands.rs
```diff
@@ -430,6 +430,24 @@ pub enum OnlineWalletSubCommand {
...
     },
+    /// Sends an original PSBT to a BIP 21 URI and broadcasts the returned Payjoin PSBT.
+    SendPayjoin {
+        /// BIP 21 URI for the Payjoin.
+        #[arg(env = "PAYJOIN_URI", long = "uri", required = true)]
+        uri: String,
+        /// URL of the Payjoin OHTTP relay. Can be repeated multiple times to attempt the
+        /// operation with multiple relays for redundancy.
+        #[arg(env = "PAYJOIN_OHTTP_RELAY", long = "ohttp_relay", required = true)]
+        ohttp_relay: Vec<String>,
```
spacebear21, line 458: nit: plural `ohttp_relays`

src/commands.rs
```diff
@@ -430,6 +430,23 @@ pub enum OnlineWalletSubCommand {
...
+        #[arg(env = "PAYJOIN_AMOUNT", long = "amount", required = true)]
+        amount: u64,
+        /// Payjoin directory which will be used to store the PSBTs which are pending action
+        /// from one of the parties.
+        #[arg(env = "PAYJOIN_DIRECTORY", long = "directory", required = true)]
+        directory: String,
+        /// URL of the Payjoin OHTTP relay. Can be repeated multiple times to attempt the
+        /// operation with multiple relays for redundancy.
+        #[arg(env = "PAYJOIN_OHTTP_RELAY", long = "ohttp_relay", required = true)]
+        ohttp_relay: Vec<String>,
```
spacebear21, line 445: nit: also plural here

src/payjoin/mod.rs
```diff
@@ -45,6 +45,646 @@ impl<'a> PayjoinManager<'a> {
...
+            &persister,
+            ohttp_keys.relay_url,
+            blockchain_client,
+        )
+        .await?;
+
+        Ok(to_string_pretty(&json!({}))?)
+    }
+
+    pub async fn send_payjoin(
```
spacebear21, line 117: This was probably meant to be in the previous commit which adds send support?

src/payjoin/mod.rs
```diff
@@ -45,6 +45,646 @@ impl<'a> PayjoinManager<'a> {
...
+    }
+
+    async fn check_proposal(
+        &mut self,
+        receiver: Receiver<UncheckedOriginalPayload>,
+        persister: &impl SessionPersister<SessionEvent = ReceiverSessionEvent>,
+        blockchain_client: BlockchainClient,
+    ) -> Result<(), Error> {
+        let next_receiver_typestate = receiver
+            .check_broadcast_suitability(None, |_| return Ok(true))
```
spacebear21: Is this intentionally hardcoded to return `true` because bdk-cli is a interactive receiver? If yes it would be clearer to use the helper `assume_interactive_receiver` to skip this validation and proceed to the next typestate.

src/payjoin/mod.rs
```diff
@@ -45,6 +45,646 @@ impl<'a> PayjoinManager<'a> {
...
+            .await
+    }
+
+    async fn apply_fee_range(
+        &mut self,
+        receiver: Receiver<WantsFeeRange>,
+        persister: &impl SessionPersister<SessionEvent = ReceiverSessionEvent>,
+        blockchain_client: BlockchainClient,
+    ) -> Result<(), Error> {
+        let next_receiver_typestate = receiver.apply_fee_range(None, None).save(persister).map_err(|e| {
```
spacebear21: Since the cli command takes a `max_fee_rate` argument we should pass that here as the `max_effective_fee_rate` option.

src/payjoin/mod.rs
```diff
@@ -45,6 +45,646 @@ impl<'a> PayjoinManager<'a> {
...
+    }
+
+    /// Syncs the blockchain once and then checks whether the Payjoin was broadcasted by the
+    /// sender.
+    ///
+    /// The currenty implementation does not support checking for the PAyjoin broadcast in a loop
+    /// and returning only when it is detected or if a timeout is reached because the [`sync_wallet`]
+    /// function consumes the BlockchainClient. BDK CLI supports multiple blockchain clients, and
+    /// at the time of writing, Kyoto consumes the client since BDK CLI is not designed for long-running
+    /// tasks.
```
Mshehu5, line 617: Is resyncing on other BlockchainClient options possible?
From the comments only kyoto is the consumes the client. Electrum, esplora and rpc are available and don't consume the client

src/payjoin/mod.rs
```diff
@@ -45,6 +45,646 @@ impl<'a> PayjoinManager<'a> {
...
+        max_fee_rate: Option<u64>,
+        ohttp_relays: Vec<String>,
+        blockchain_client: BlockchainClient,
+    ) -> Result<String, Error> {
+        let address = self
+            .wallet
+            .next_unused_address(bdk_wallet::KeychainKind::External);
+
+        let ohttp_relays: Vec<url::Url> = ohttp_relays
+            .into_iter() // if ohttp_relay: Option<Vec<String>>
```
Mshehu5: Niit: unused comment removal

src/commands.rs
mehmetefeumit: The reason I kept this singular here is that it works better from a CLI options perspective:

```
send_payjoin --fee_rate 1 --ohttp_relay "https://pj.bobspacebkk.com" --ohttp_relay "https://ohttp.achow101.com" --uri "..."
```

Having the options be `--ohttp_relays` would not be accurate. And to my knowledge there isn't another way to past that list in the command line.

src/commands.rs
mehmetefeumit: Same comment as the above.

The reason I kept this singular here is that it works better from a CLI options perspective:

```
send_payjoin --fee_rate 1 --ohttp_relay "https://pj.bobspacebkk.com" --ohttp_relay "https://ohttp.achow101.com" --uri "..."
```

Having the options be `--ohttp_relays` would not be accurate. And to my knowledge there isn't another way to past that list in the command line.

src/payjoin/mod.rs
mehmetefeumit: Whoops. Fixing.

Cargo.toml
mehmetefeumit: I am changing the function signatures to use payjoin::IntoUrl, but going to hold on to the Url dependency to parse the URLs before passing them to the Payjoin process functions. Let me know if there is an alternative approach 👍🏻

src/payjoin/mod.rs
mehmetefeumit: See my comment below on the main discussion.

src/commands.rs
spacebear21: Fair enough, I guess I was thinking about it from the perspective of a config file but I realize now this only supports command line argument parsing so it makes sense.

src/payjoin/mod.rs
```diff
@@ -167,6 +239,508 @@ impl<'a> PayjoinManager<'a> {
...
+        );
+        return self
+            .monitor_payjoin_proposal(next_receiver_typestate, persister, blockchain_client)
+            .await;
+    }
+
+    /// Syncs the blockchain once and then checks whether the Payjoin was broadcasted by the
+    /// sender.
+    ///
+    /// The currenty implementation does not support checking for the PAyjoin broadcast in a loop
```
arminsabouri: nit: PAyjoin -> Payjoin

mehmetefeumit:
@notmandatory I was able to reproduce the error on my local. The issue was the incorrect use of the tokio tick function, as it returns immediately in the first call (see example [here](https://docs.rs/tokio/1.48.0/tokio/time/struct.Interval.html#method.tick). I replaced it with a sleep as it is more appropriate for the scenario, and made sure that it is before the sync so that the receiver gives the sender 3 seconds before syncing and checking.

Secondly, the error you got should have been handled. It indicates that the transaction has not been broadcast; the receiver does not have the transaction in their indexed graph. This is because the receiver did not create the transaction, and signing it does not add the transaction to the indexed graph of the wallet. The receiver should sync the blockchain to determine a new transaction which belongs to them.

I thought about whether the transaction should be added to the wallet graph at the time of signing, and I believe it should not (which is the current implementation). If the sender adds non-SegWit inputs, then the TXID will change and that entry in the receiver wallet graph will be left dangling.

TL;DR: Ready for review :D

src/payjoin/mod.rs
```diff
@@ -0,0 +1,858 @@
...
+        let address = self
+            .wallet
+            .next_unused_address(bdk_wallet::KeychainKind::External);
+
+        let ohttp_relays: Vec<url::Url> = ohttp_relays
+            .into_iter()
+            .map(|s| url::Url::parse(&s))
+            .collect::<Result<_, _>>()
+            .map_err(|e| {
+                Error::Generic(format!("Failed to parse one or more OHTTP URLs: {}", e))
```
tvpeter: nit: For consistency, can you pass the error directly like:

`Error::Generic(format!("Failed to parse one or more OHTTP URLs: {e}")`

src/handlers.rs
```diff
@@ -1325,6 +1221,170 @@ async fn respond(
...
 }
 
+#[cfg(any(
+    feature = "electrum",
+    feature = "esplora",
+    feature = "cbf",
+    feature = "rpc"
+))]
+/// Syncs a given wallet using the blockchain client.
+pub async fn sync_wallet(client: BlockchainClient, wallet: &mut Wallet) -> Result<(), Error> {
```
tvpeter, line 1231: nit: Will be better to split them into different modules, but don't worry about it for now.

tvpeter (approved):
tACK c6bce58871e23c26d116e5c7b767ff4209a7b3c4

Thank you for working on this once again @mehmetefeumit .

PR #203, opened 2025-06-18
Add initializing wallet configuration with Justfile

### LINKED ISSUE #192: Add "init" wallet support
**Describe the enhancement**  

Currently all parameters need to be specified in env variables or as cli options for every request. With so many options this is hard to setup and use. It would be nice if a user could "initialize" a wallet name with all the parameters they need, such as descriptors, network, db, server, etc. 

Maybe this could be done with `Justfile` that set the env variables, in a toml file, or in some sort of db file.

**Use case**  

I want to initialize a Testnet wallet with two descriptors, a sqlite db file, and local regtest server url and authentication. Id like to store all these parameters once with my wallet name and then use them every time I specify that wallet name. 

**Additional context**

Something like this is done with the `bdk` cli examples: https://github.com/bitcoindevkit/bdk/blob/542f9675c0716dfc5b4f3ab4e5320702e9f07a37/examples/example_cli/src/lib.rs#L72

tvpeter:
### Description

This PR adds `just init` command to set wallet parameters. 

Fixes #192 

### Notes to the reviewers

- Reusing the exported `serde` crate from `bdk_wallet` did not offer the `derive` feature
-  Set the default value for `database_type` to `sqlite`
-  CLI parameters have higher precedence over saved config values except for the database_type as it has a default value

#### All Submissions:

#### New Features:


notmandatory:
I like the approach for loading the config file parameters if the file exists, but is there a reason you didn't add a CLI `bdk-cli init ...` command in rust to create the config file rather than using `just` ? 

Creating the config file in rust should be easier to maintain/keep in sync than your `just` approach and shouldn't take too much more than adding the `Serialize` trait to `WalletConfigInner`. You could possibly reuse the `WalletOpts` to capture the values with `clap` too.

tvpeter:
> I like the approach for loading the config file parameters if the file exists, but is there a reason you didn't add a CLI `bdk-cli wallet init ...` command in rust to create the config file rather than using `just` ?
> 
> Creating the config file in rust should be easier to maintain/keep in sync than your `just` approach and shouldn't take too much more than adding the `Serialize` trait to `WalletConfigInner`. You should be able to reuse the `WalletOpts` to capture the values with `clap` that you need to put in the config file too.

Alright, I will update.

tvpeter:
> Alright, I will update.

@notmandatory I have updated the PR

notmandatory:
I spent some time today reviewing and even though this looks like a workable way to do it and is based on my suggestion I'm afraid it's going to be a hassle to maintain.  I've been experimenting with somehow using the different clap parsing functions (https://docs.rs/clap/latest/clap/trait.Parser.html) but so far haven't figured out a better way. I'd like to keep thinking about it. We might have to simplify the problem somehow such as by not trying to merge loaded and CLI args.

tvpeter:
>  I'm afraid it's going to be a hassle to maintain. I've been experimenting with somehow using the different clap parsing functions (https://docs.rs/clap/latest/clap/trait.Parser.html) but so far haven't figured out a better way. I'd like to keep thinking about it. We might have to simplify the problem somehow such as by not trying to merge loaded and CLI args.

Thank you @notmandatory.
Yes, I agree that it will be challenging to maintain, as it involves two steps (pre clap parsing and after parsing) and even more difficult as they are in different locations in the codebase.  I thought about moving everything to pre clap parsing based on the wallet sub-command the user has entered but that will clog the entry point as there will be lots of `if...else` checks. 

Also, for the Parser trait, I read that it does not allow interacting with an external resource such as reading a file as I would want it. 

I will also be trying out other approaches in case there is a cleaner way to handle it.

notmandatory:
How about something like:

1. One command to save a wallet config, give wallet name or a default name is used, clap enforces required options; optional -f to override existing config:
   ```wallet [-f] [-w <name>] config <wallet opts>```
2. All other wallet commands require that a config exists, use wallet name or the default is used: 
    ```wallet [-w <name>] sync | balance | new_address | etc...```
3.  Repl uses the same wallet configs:
    ```repl [-w <name>]```
 4. Should throw warnings if using mainnet and saving a private descriptor in a config.

tvpeter:
> How about something like:
> 
> 1. One command to save a wallet config, give wallet name or a default name is used, clap enforces required options; optional -f to override existing config:
>    `wallet [-f] [-w <name>] config <wallet opts>`
> 2. All other wallet commands require that a config exists, use wallet name or the default is used. Give error if config is missing:
>    `wallet [-w <name>] sync | balance | new_address | etc...`
> 3. Repl uses the same wallet configs:
>    `repl [-w <name>]`
> 4. Add a command to list all available config names with summary of configured options:
>    `wallet list`
> 5. Should throw warnings if using mainnet and saving a private descriptor in a config.

Hi @notmandatory, there is an issues that we did not envisage and discuss on this PR that came up during implementation:
The `wallet [-w <name>] config <wallet opts>` command introduced a `wallet name` parameter that has to be supplied whenever the wallet command is used. This means that the `list` subcommand that should not require a wallet name has to accept the wallet name. If we make the wallet name optional above, we will have to do manual validation for the name.

README.md
```diff
@@ -202,3 +202,44 @@ You can optionally return outputs of commands in  human-readable, tabular format
...
+
+To overwrite an existing wallet configuration, use the  `--force` flag after the `config` sub-command.
+
+#### Using a Configured Wallet
+
+Once configured, use any wallet command with just the wallet name:
+
+
+```shell
+cargo run --features electrum wallet -w my_wallet  new_address
```
notmandatory: ⛏️ nit, has extra space:
```suggestion
cargo run --features electrum wallet -w my_wallet new_address
```

src/commands.rs
```diff
@@ -156,14 +169,15 @@ pub enum ClientType {
 #[derive(Debug, Args, Clone, PartialEq, Eq)]
 pub struct WalletOpts {
     /// Selects the wallet to use.
-    #[arg(env = "WALLET_NAME", short = 'w', long = "wallet")]
+    #[arg(skip)]
     pub wallet: Option<String>,
+    // #[arg(env = "WALLET_NAME", short = 'w', long = "wallet", required = true)]
```
notmandatory: Commented out line should be removed.
```suggestion
```

src/utils.rs
```diff
@@ -360,7 +344,39 @@ pub async fn sync_kyoto_client(wallet: &mut Wallet, client: Box<LightClient>) ->
...
+        .get(wallet_name)
+        .ok_or(Error::Generic(format!(
+            "Wallet '{wallet_name}' not found in config"
+        )))?;
+
+    let network = match wallet_config.network.as_str() {
+        "bitcoin" => Ok(Network::Bitcoin),
+        "testnet" => Ok(Network::Testnet),
+        "regtest" => Ok(Network::Regtest),
+        "signet" => Ok(Network::Signet),
```
notmandatory, line 644: 🔨 need to add Testnet4:
```suggestion
        "signet" => Ok(Network::Signet),
        "testnet4" => Ok(Network::Testnet4),
```

thunderbiscuit (commented):
Thanks for this! I'll use it for sure once it gets in. I have two little things I want to ask, take them or leave them they're not blockers at all.

## Question

When I run the example in the readme, I get the following config file:

```toml
network = "signet"

[wallets.my_wallet]
wallet = "my_wallet"
network = "signet"
ext_descriptor = "tr(tprv8Z.../0/*)#dtdqk3dx"
int_descriptor = "tr(tprv8Z.../1/*)#ulgptya7"
database_type = "sqlite"
client_type = "electrum"
server_url = "ssl://mempool.space:60602"
batch_size = 10
```

The network is somehow added as a top-level key, but I suspect it's meant to only be added under the specific wallet config maybe? It doesn't prevent the cli from working as expected however so all is good there.

## Comment

I always find it hard when applications require a config file but it's empty by default. I have to search in their docs to find what's available, and even then it's hard to know if you miss anything. A nicer approach is when the config file comes pre-populated, with all keys commented out to start with but they're at least all there so I can use the config file to sort of "discover" what's available in the wallet. This is not always possible for all keys, but could be cool here. When doing this you could do it 2 ways: 
1. Your config file could just come pre-made with a top-level commented section that has all keys available and the rest works as it currently does (easiest to implement). Note that you could also just add this as a codeblock in the readme (a complete list of available config options).
2. Each newly created wallet comes with its own section of complete config keys, some active and the rest commented out.

tvpeter:
> Thanks for this! I'll use it for sure once it gets in. I have two little things I want to ask, take them or leave them they're not blockers at all.
> 
> ## Question
> When I run the example in the readme, I get the following config file:
> 
> ```toml
> network = "signet"
> 
> [wallets.my_wallet]
> wallet = "my_wallet"
> network = "signet"
> ext_descriptor = "tr(tprv8Z.../0/*)#dtdqk3dx"
> int_descriptor = "tr(tprv8Z.../1/*)#ulgptya7"
> database_type = "sqlite"
> client_type = "electrum"
> server_url = "ssl://mempool.space:60602"
> batch_size = 10
> ```
> 
> The network is somehow added as a top-level key, but I suspect it's meant to only be added under the specific wallet config maybe? It doesn't prevent the cli from working as expected however so all is good there.
> 
> ## Comment
> I always find it hard when applications require a config file but it's empty by default. I have to search in their docs to find what's available, and even then it's hard to know if you miss anything. A nicer approach is when the config file comes pre-populated, with all keys commented out to start with but they're at least all there so I can use the config file to sort of "discover" what's available in the wallet. This is not always possible for all keys, but could be cool here. When doing this you could do it 2 ways:
> 
> 1. Your config file could just come pre-made with a top-level commented section that has all keys available and the rest works as it currently does (easiest to implement). Note that you could also just add this as a codeblock in the readme (a complete list of available config options).
> 2. Each newly created wallet comes with its own section of complete config keys, some active and the rest commented out.

Thank you for the review @thunderbiscuit 
I will include this feedback in the next update as I also look at the [configure_me](https://docs.rs/configure_me/latest/configure_me) lib to pick improvements.

tvpeter:
> as I also look at the [configure_me](https://docs.rs/configure_me/latest/configure_me) lib to pick improvements

I have done a quick review of configure_me, and from my perspective, it may not be suitable for the CLI's use case. It uses a declarative approach to define the config file and generates a static file that updating it requires manually editing and recompiling. It is more suitable for an application-wide configuration and will be challenging to use for a per-wallet configuration. Also, I have not seen how it can handle conditional features. Lastly, I am not sure it can integrate with clap as they seemed to serve the same purpose (might be wrong).  But there are improvements that we can pick from the library, things like validation and maybe how we merge fetched values into the CLI values.

notmandatory:
thanks for looking into it, I agree we should stick with our simpler toml file approach.

src/config.rs
```diff
@@ -80,88 +82,181 @@ impl WalletConfig {
...
+            "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/0/*)#429nsxmg"
+        );
+
+        #[cfg(any(
+            feature = "electrum",
+            feature = "esplora",
+            feature = "rpc",
+            feature = "cbf"
+        ))]
+        assert_eq!(opts.client_type, ClientType::Esplora);
```
nymius, line 211: While building `bdk-cli` inside a flake, I've got the following error:
```bash
> cargoCheckHook flags: -j 16 --profile release --features=repl\,sqlite\,rpc\,sp --target x86_64-unknown-linux-gnu --offline --
>    Compiling bdk-cli v2.0.0 (/build/gbaaz3igwf8dv2785f5vzxhnmhh5g41g-source)
> error[E0599]: no variant or associated item named `Esplora` found for enum `commands::ClientType` in the current scope
>    --> src/config.rs:211:50
>     |
> 211 |         assert_eq!(opts.client_type, ClientType::Esplora);
>     |                                                  ^^^^^^^ variant or associated item not found in `commands::ClientType`
>     |
>    ::: src/commands.rs:188:1
>     |
> 188 | pub enum ClientType {
>     | ------------------- variant or associated item `Esplora` not found for this enum
>
```

I'm not fully aware of what was being tested here, but this is causing cross incompatibilities with the features enabled for the enum and the features enabled for the assert.

tvpeter: The issue is that the `Esplora` client was compiled out, but was used in the config test as the client for any enabled client type. 

Thank you for catching this just in time

PR #207, opened 2025-07-21
feat(hwi): add hardware wallet support

### LINKED ISSUE #194: Add back HWI signing support
**Describe the enhancement**  

As part of the 1.0 update we removed HWI signing support. The original support was using the python based HWI library. We should use the Wizardsardine async-hwi tool to re-add this capability. 

**Use case**  

This will provide an example for how to integrate `bdk_wallet` with the `async-hwi` library and allow bdk-cli users to experiment and test with different hardware signers that it supports.

**Additional context**

https://github.com/wizardsardine/async-hwi

tvpeter:
### Description
fixes #194 

### Notes to the reviewers

- enable offline signer for creating tx whenever hwi feature is enabled
- make handle_offline_wallet_subcommand fn async to handle hwi operations
- The `display_address` method in the provided trait in async-hwi library is meant to trigger address generation by the connected device and display it on the device screen. Generated addresses are the same for both the app and device, so app addresses are returned for the `hwi address` subcommand.

This feature was tested with `SpecterSimulator` and `Coldcard`, although not all `hwi` features were successfully tested on coldcard.

For signing with hwi, it does sign but do not finalized the `PSBT`, so it has to be finalized before broadcasting. 

### Checklists

HWI Wallet Subcommands implementation
* [x] hwi devices
* [x] hwi register
* [x] hwi address
* [x] hwi sign tx

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -23,6 +23,11 @@ use clap::{Args, Parser, Subcommand, ValueEnum, value_parser};
 use crate::utils::parse_proxy_auth;
 use crate::utils::{parse_address, parse_outpoint, parse_recipient};
 
+#[cfg(feature = "hwi")]
+const ENABLE_OFFLINE_SIGNER: bool = true;
+#[cfg(not(feature = "hwi"))]
+const ENABLE_OFFLINE_SIGNER: bool = false;
+
 /// The BDK Command Line Wallet App
 ///
 /// bdk-cli is a lightweight command line bitcoin wallet, powered by BDK.
@@ -305,7 +310,7 @@ pub enum OfflineWalletSubCommand {
         #[arg(long = "enable_rbf", short = 'r', default_value_t = true)]
         enable_rbf: bool,
         /// Make a PSBT that can be signed by offline signers and hardware wallets. Forces the addition of `non_witness_utxo` and more details to let the signer identify the change output.
-        #[arg(long = "offline_signer")]
+        #[arg(long = "offline_signer", default_value_t=ENABLE_OFFLINE_SIGNER)]
         offline_signer: bool,
         /// Selects which utxos *must* be spent.
         #[arg(env = "MUST_SPEND_TXID:VOUT", long = "utxos", value_parser = parse_outpoint)]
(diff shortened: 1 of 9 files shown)


src/utils.rs
```diff
@@ -631,3 +642,111 @@ pub fn format_descriptor_output(result: &Value, pretty: bool) -> Result<String,
 
     Ok(format!("{table}"))
 }
+
+#[cfg(feature = "hwi")]
+pub async fn connect_to_hardware_wallet(
```
sdmg15: I feel like the function `connect_to_hardware_wallet` should be named `get_connected_hardware_devices` instead ?
That's according to the purpose of the subcommand `HwiSubCommand::Devices` (returning the list of connected hardware devices).
If it's assumed the user will always have only one connected device then I'd suggest renaming the variant to just `HwiSubCommand::Device`

src/commands.rs
```diff
@@ -228,6 +241,23 @@ pub struct WalletOpts {
...
+#[cfg(feature = "hwi")]
+#[derive(Clone, Debug, PartialEq, Eq, Args)]
+pub struct HwiOpts {
+    /// Wallet name
+    #[arg(env = "WALLET", short = 'w', long = "wallet")]
+    pub wallet: Option<String>,
+
+    /// External descriptor
+    #[arg(env = "EXT_DESCRIPTOR", short = 'e', long = "ext_descriptor")]
+    pub ext_descriptor: Option<String>,
```
sdmg15, line 268: I  noticed some inconsistency with the option `ext-descriptor`, some commands uses `ext-descriptor` but the added feature uses `ext_descriptor`.  Same for `--database_type`.

notmandatory (approved):
utACK 8a6012d27980f92e01bde5e39d78b96952df42a0

Looks like last commit failed due to some unrelated CI/codecov error.

notmandatory:
Thanks for getting this one done, it's a really important feature! We should talk this one up when the next release is published.

tvpeter:
> Looks like last commit failed due to some unrelated CI/codecov error.

Yes, coveralls.io is having a DNS error
Screenshot 2026-02-26 at 12 07 17 AM

tvpeter:
Closing this PR in favour of #311

PR #212, opened 2025-08-18
feat(pretty-format): add pretty flag to format outputs in a tabular form

### LINKED ISSUE #193: Add option to return command results in "pretty" format
**Describe the enhancement**  

Currently all command outputs are in JSON format. It would be nice to have an option to return the outputs in a more human readable format, such as formatted tables or similar

This could be done with simple println and tabs or with a TUI library that provides better formatting options.

**Use case**  

Json is good for the default but for casual testing or new users a human readable format is better. 

**Additional context**

This is an example of prettier table format for TUI output:

https://github.com/ccbrown/iocraft?tab=readme-ov-file#more-examples

tvpeter:
Fixes #193 

### Description

This PR adds a top level flag `--pretty` to format output in a tabular form to enhance readability for users.

## Changelog notice
- Add `--pretty` top level flag for formatting commands output in a tabular format

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ page. See [DEVELOPMENT_CYCLE.md](DEVELOPMENT_CYCLE.md) for more details.
 ## [Unreleased]
 
 - Removed MSRV and bumped Rust Edition to 2024
+- Add `--pretty` top level flag for formatting commands output in a tabular format
 
 ## [1.0.0]
 
--- a/README.md
+++ b/README.md
@@ -193,3 +193,12 @@ Note: You can modify the `Justfile` to reflect your nodes' configuration values.
    cargo run --features rpc -- wallet -u "127.0.0.1:18443" -c rpc -a user:password sync
    cargo run --features rpc -- wallet -u "127.0.0.1:18443" -c rpc -a user:password balance
    ```
+
+## Formatting Responses using `--pretty` flag
+
+You can optionally return outputs of commands in  human-readable, tabular format instead of `JSON`. To enable this option, simply add the `--pretty` flag as a top level flag. For instance, you wallet's balance in a pretty format, you can run:
+
+```shell
+cargo run --pretty -n signet wallet -w {wallet_name} -d sqlite balance
+```
+This is available for wallet, key, repl and compile features. When ommitted, outputs default to `JSON`.
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -22,11 +22,11 @@ use crate::utils::*;
 use bdk_redb::Store as RedbStore;
 use bdk_wallet::bip39::{Language, Mnemonic};
 use bdk_wallet::bitcoin::{
+    Address, Amount, FeeRate, Network, Psbt, Sequence, Txid,
     bip32::{DerivationPath, KeySource},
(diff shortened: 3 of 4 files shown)


notmandatory:
I manually tested and everything works fine, my only small nit is that the `wallet unspent` command is too long to fit on one line for even my wide monitor. But I don't no a better way to pretty display everything without somehow abbreviating some of the column data. Would it make sense do apply a function like this on the outpoint and block hash cells, and it would also be a little smaller and more useful to display the Output ScriptPubkey as an address rather than as the raw script bytes. 

Something like this in `utils.rs`:

```rust
pub(crate) fn shorten(displayable: impl Display, start: u8, end: u8) -> String {
    let displayable = displayable.to_string();
    let start_str: &str = &displayable[0..start as usize];
    let end_str: &str = &displayable[displayable.len() - end as usize..];
    format!("{start_str}...{end_str}")
}
```

Then the "outputs" match arm would look like:
```rust
Unspent => {
            let utxos = wallet.list_unspent().collect::<Vec<_>>();
            if cli_opts.pretty {
                let mut rows: Vec<Vec<CellStruct>> = vec![];
                for utxo in &utxos {
                    let height = utxo
                        .chain_position
                        .confirmation_height_upper_bound()
                        .map(|h| h.to_string())
                        .unwrap_or("Pending".to_string());

                    let block_hash = match &utxo.chain_position {
                        ChainPosition::Confirmed { anchor, .. } => anchor.block_id.hash.to_string(),
                        ChainPosition::Unconfirmed { .. } => "Unconfirmed".to_string(),
                    };

                    rows.push(vec![
                        shorten(utxo.outpoint, 8, 10).cell(),
                        utxo.txout
                            .value
                            .to_sat()
                            .to_string()
                            .cell()
                            .justify(Justify::Right),
                        Address::from_script(&utxo.txout.script_pubkey, cli_opts.network).unwrap().cell(),
                        utxo.keychain.cell(),
                        utxo.is_spent.cell(),
                        utxo.derivation_index.cell(),
                        height.to_string().cell().justify(Justify::Right),
                        shorten(block_hash,8,8).cell().justify(Justify::Right),
                    ]);
                }
                let table = rows
                    .table()
                    .title(vec![
                        "Outpoint".cell().bold(true),
                        "Output (sat)".cell().bold(true),
                        "Output Address".cell().bold(true),
                        "Keychain".cell().bold(true),
                        "Is Spent".cell().bold(true),
                        "Index".cell().bold(true),
                        "Block Height".cell().bold(true),
                        "Block Hash".cell().bold(true),
                    ])
                    .display()
                    .map_err(|e| Error::Generic(e.to_string()))?;
                Ok(format!("{table}"))
            } else {
                Ok(serde_json::to_string_pretty(&utxos)?)
            }
        }
```

tvpeter:
> I manually tested and everything works fine, my only small nit is that the `wallet unspent` command is too long to fit on one line for even my wide monitor. But I don't know a better way to pretty display everything without somehow abbreviating some of the column data. I think it would make sense to apply a function on the outpoint and block hash cells, and it would also be a little smaller and more useful to display the Output ScriptPubkey as an address rather than as the raw script bytes.

Updated. Thank you for the suggestion, it fits just great.

PR #220, opened 2025-09-10
Add experimental silent payment transaction creation support

nymius:
**Description**

This PR adds experimental support for creating silent payment transactions through a new `CreateSpTx` command. The implementation integrates the `bdk_sp` crate to enable sending Bitcoin to silent payment addresses.

Key changes:
- Adds `bdk_sp` dependency as an optional feature
- Implements `CreateSpTx` command with support for silent payment recipients
- Includes parser for silent payment `address:amount` pairs
- Returns signed transactions ready for broadcast (not PSBTs due to silent payment derivation constraints)

**Notes to the reviewers**

- This feature is marked as **EXPERIMENTAL** and includes warnings against mainnet use
- The command returns signed transactions directly rather than PSBTs because silent payment script pubkey derivation cannot be performed securely in a trustless manner with standard PSBT workflows
- RBF is disabled for silent payment transactions (sequence set to MAX)
- The implementation handles both single key and extended key signers
- Error handling uses temporary `.expect()` calls that should be addressed in future iterations

**Changelog notice**

**Added**: Experimental silent payment transaction creation via `CreateSpTx` command (feature-gated behind `sp` flag)

**Checklists**

**All Submissions**

* [x] I followed the [conventional commit guidelines](https://www.conventionalcommits.org/en/v1.0.0/)
* [x] I ran cargo fmt, clippy and test before committing

**New Features**

~~* [ ] I've added tests for the new feature~~

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -13,6 +13,9 @@
 //! All subcommands are defined in the below enums.
 
 #![allow(clippy::large_enum_variant)]
+#[cfg(feature = "silent-payments")]
+use {crate::utils::parse_sp_code_value_pairs, bdk_sp::encoding::SilentPaymentCode};
+
 use bdk_wallet::bitcoin::{
     Address, Network, OutPoint, ScriptBuf,
     bip32::{DerivationPath, Xpriv},
@@ -181,6 +184,19 @@ pub enum CliSubCommand {
         /// Target shell syntax
         #[arg(value_enum)]
         shell: Shell,
+    },
+    /// Silent payment code generation tool.
+    ///
+    /// Allows the encoding of two public keys into a silent payment code.
+    /// Useful to create silent payment transactions using fake silent payment codes.
+    #[cfg(feature = "silent-payments")]
+    SilentPaymentCode {
+        /// The scan public key to use on the silent payment code.
+        #[arg(long = "scan_key")]
+        scan: bdk_sp::bitcoin::secp256k1::PublicKey,
+        /// The spend public key to use on the silent payment code.
+        #[arg(long = "spend_key")]
+        spend: bdk_sp::bitcoin::secp256k1::PublicKey,
     },
 }
 
@@ -360,6 +376,62 @@ pub enum OfflineWalletSubCommand {
         /// Enables Replace-By-Fee (BIP125).
         #[arg(long = "enable_rbf", short = 'r', default_value_t = true)]
         enable_rbf: bool,
+        /// Make a PSBT that can be signed by offline signers and hardware wallets. Forces the addition of `non_witness_utxo` and more details to let the signer identify the change output.
+        #[arg(long = "offline_signer")]
+        offline_signer: bool,
+        /// Selects which utxos *must* be spent.
+        #[arg(env = "MUST_SPEND_TXID:VOUT", long = "utxos", value_parser = parse_outpoint)]
+        utxos: Option<Vec<OutPoint>>,
+        /// Marks a utxo as unspendable.
+        #[arg(env = "CANT_SPEND_TXID:VOUT", long = "unspendable", value_parser = parse_outpoint)]
+        unspendable: Option<Vec<OutPoint>>,
+        /// Fee rate to use in sat/vbyte.
+        #[arg(env = "SATS_VBYTE", short = 'f', long = "fee_rate")]
+        fee_rate: Option<f32>,
+        /// Selects which policy should be used to satisfy the external descriptor.
+        #[arg(env = "EXT_POLICY", long = "external_policy")]
+        external_policy: Option<String>,
+        /// Selects which policy should be used to satisfy the internal descriptor.
+        #[arg(env = "INT_POLICY", long = "internal_policy")]
+        internal_policy: Option<String>,
+        /// Optionally create an OP_RETURN output containing given String in utf8 encoding (max 80 bytes)
+        #[arg(
+            env = "ADD_STRING",
+            long = "add_string",
+            short = 's',
+            conflicts_with = "add_data"
+        )]
+        add_string: Option<String>,
+        /// Optionally create an OP_RETURN output containing given base64 encoded String. (max 80 bytes)
+        #[arg(
+            env = "ADD_DATA",
+            long = "add_data",
+            short = 'o',
+            conflicts_with = "add_string"
+        )]
+        add_data: Option<String>, //base 64 econding
+    },
+    /// Creates a silent payment transaction
+    ///
+    /// This sub-command is **EXPERIMENTAL** and should only be used for testing. Do not use this
+    /// feature to create transactions that spend actual funds on the Bitcoin mainnet.
+
+    // This command DOES NOT return a PSBT. Instead, it directly returns a signed transaction
+    // ready for broadcast, as it is not yet possible to perform a shared derivation of a silent
+    // payment script pubkey in a secure and trustless manner.
+    #[cfg(feature = "silent-payments")]
+    CreateSpTx {
+        /// Adds a recipient to the transaction.
+        // Clap Doesn't support complex vector parsing https://github.com/clap-rs/clap/issues/1704.
+        // Address and amount parsing is done at run time in handler function.
+        #[arg(env = "ADDRESS:SAT", long = "to", required = false, value_parser = parse_recipient)]
+        recipients: Option<Vec<(ScriptBuf, u64)>>,
+        /// Parse silent payment recipients
+        #[arg(long = "to-sp", required = true, value_parser = parse_sp_code_value_pairs)]
+        silent_payment_recipients: Vec<(SilentPaymentCode, u64)>,
+        /// Sends all the funds (or all the selected utxos). Requires only one recipient with value 0.
+        #[arg(long = "send_all", short = 'a')]
+        send_all: bool,
         /// Make a PSBT that can be signed by offline signers and hardware wallets. Forces the addition of `non_witness_utxo` and more details to let the signer identify the change output.
         #[arg(long = "offline_signer")]
         offline_signer: bool,
--- a/src/error.rs
+++ b/src/error.rs
@@ -21,6 +21,10 @@ pub enum BDKCliError {
     #[error("Create transaction error: {0}")]
     CreateTx(#[from] bdk_wallet::error::CreateTxError),
 
+    #[cfg(feature = "silent-payments")]
+    #[error("Silent payment address decoding error: {0}")]
+    SilentPaymentParseError(#[from] bdk_sp::encoding::ParseError),
+
     #[error("Descriptor error: {0}")]
     DescriptorError(#[from] bdk_wallet::descriptor::error::Error),
 
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -49,6 +49,15 @@ use bdk_wallet::{
 use clap::CommandFactory;
 use cli_table::{Cell, CellStruct, Style, Table, format::Justify};
 use serde_json::json;
+#[cfg(feature = "silent-payments")]
+use {
+    bdk_sp::{
+        bitcoin::{PrivateKey, PublicKey, ScriptBuf},
+        encoding::SilentPaymentCode,
+        send::psbt::derive_sp,
+    },
+    bdk_wallet::keys::{DescriptorPublicKey, DescriptorSecretKey, SinglePubKey},
+};
 
 #[cfg(feature = "electrum")]
 use crate::utils::BlockchainClient::Electrum;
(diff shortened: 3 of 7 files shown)


src/handlers.rs
```diff
@@ -318,7 +327,152 @@ pub fn handle_offline_wallet_subcommand(
...
+            }
+
+            if let Some(fee_rate) = fee_rate {
+                if let Some(fee_rate) = FeeRate::from_sat_per_vb(fee_rate as u64) {
+                    tx_builder.fee_rate(fee_rate);
+                }
+            }
+
+            if let Some(utxos) = utxos {
+                tx_builder.add_utxos(&utxos[..]).unwrap();
```
tvpeter: Can you use the Error enum from the error module to gracefully handle errors rather than unwrapping?

src/utils.rs
```diff
@@ -51,6 +53,25 @@ pub(crate) fn parse_recipient(s: &str) -> Result<(ScriptBuf, u64), String> {
     Ok((addr.script_pubkey(), val))
 }
 
+#[cfg(feature = "sp")]
+pub(crate) fn parse_sp_code_value_pairs(s: &str) -> Result<(SilentPaymentCode, u64), String> {
```
tvpeter: you can return the Error and use the Generic variant instead of String here too.

src/commands.rs
```diff
@@ -315,6 +318,62 @@ pub enum OfflineWalletSubCommand {
...
+    // This command DOES NOT return a PSBT. Instead, it directly returns a signed transaction
+    // ready for broadcast, as it is not yet possible to perform a shared derivation of a silent
+    // payment script pubkey in a secure and trustless manner.
+    #[cfg(feature = "sp")]
+    CreateSpTx {
+        /// Adds a recipient to the transaction.
+        // Clap Doesn't support complex vector parsing https://github.com/clap-rs/clap/issues/1704.
+        // Address and amount parsing is done at run time in handler function.
+        #[arg(env = "ADDRESS:SAT", long = "to", required = false, value_parser = parse_recipient)]
+        recipients: Option<Vec<(ScriptBuf, u64)>>,
```
tvpeter, line 428: is there a need to add these recipients again since the focus is on sp_recipients?

src/commands.rs
```diff
@@ -315,6 +318,62 @@ pub enum OfflineWalletSubCommand {
...
+        /// Parse silent payment recipients
+        #[arg(long = "to-sp", required = true, value_parser = parse_sp_code_value_pairs)]
+        silent_payment_recipients: Vec<(SilentPaymentCode, u64)>,
```
tvpeter, line 431: how can one construct sp_recipients? can you add that to the readme?

tvpeter (changes requested):
Thank you for working on adding this feature @nymius,

I couldn't test because I don't know how to construct sp recipients. I will check the documentation on the library and will come back to test it. In the meantime, I have left some comments.
Thank you

src/commands.rs
nymius: I want to show that's still possible to send a non silent payment output together with a silent payment output in the same transaction, that's why I left this here.

src/commands.rs
nymius: Each silent payment recipient is a string of the form `<silent-payment-address>:<amount>`, for example: `sprt1qq0u4yswlkqx36shz7j8mwt335p4el5txc8tt6yny3dqewlw4rwdqkqewtzh728u7mzkne3uf0a35mzqlm0jf4q2kgc5aakq4d04a9l734u5ddn6e:1000` (this is a regtest address, because human readable prefix is `sprt`).
You cannot get addresses from your `bdk-cli` because I don't think the receiving side is mature enough yet, but;
`SilentPaymentCode` is prepared to decode any silent payment encoded address, so I recommend you to take a look in [`bdk-sp` encoding tests](https://github.com/bitcoindevkit/bdk-sp/blob/master/silentpayments/src/encoding/mod.rs#L392) to get some fake addresses from there.
If you want to create silent payment addresses dynamically you would have to compile `sp-cli2`, the cli implemented in `bdk-sp`, maybe that's too much to add to the `README` (to ask to compile another tool to get one), what do you think?

tvpeter: What I think is that, if it is possible, you should add deriving SP addresses to this PR. I wanted to test how you implemented it in the CLI v2 but it broke. Else you can add a link to the test addresses so it is easy for users to find them.

src/handlers.rs
nymius: I'm reviewing this unwraps again. I was basing these changes in the `CreateTx` command, that has the same number of unwraps on the same places.
I can come up with a graceful error handling for this PR, but then we would need to open a new PR to update the other command.
Also, as I'm using the unwraps on the same places, then the Error variants should be as general as possible to reuse them in the `CreateTx` command, and then this PR wouldn't be doing a single thing, but two.
Any recommendations about how to proceed?

src/utils.rs
nymius: Same comment than above, this function is following the style of `parse_recipient`. The error should be generic in terms of parsing or should be particular to silent payments?

src/commands.rs
nymius: But if we add the chances to derive addresses, then we are allowing the creation of transactions locking funds into those addresses. I could restrict this functionality to only derive testnet addresses, so no one lose any real funds.
I mean, is problematic to open the door to receive without implementing the whole functionality to scan for new outputs.

src/handlers.rs
tvpeter: Yeah, I have noticed that the `CreateTx` command has unwrappings, which we have not updated after adding the Error enum. If you can come up with variants to handle both cases, that would be great. After you have handled the `CreateSpTx` case, you can add an extra commit to fix the case for `CreateTx`.

src/utils.rs
tvpeter: if your library is returning a specific error for this, you can add it to the enum or you can just use the `generic` variant

src/commands.rs
tvpeter: Yes, only testnet addresses are fine. We generally warn users against using this tool for mainnet.

src/error.rs
```diff
@@ -21,6 +21,9 @@ pub enum BDKCliError {
     #[error("Create transaction error: {0}")]
     CreateTx(#[from] bdk_wallet::error::CreateTxError),
 
+    #[error("Silent payment address decoding error: {0}")]
```
tvpeter, line 25:
```suggestion
    #[cfg(feature = "sp")]
    #[error("Silent payment address decoding error: {0}")]
```

src/handlers.rs
```diff
@@ -323,7 +333,160 @@ pub fn handle_offline_wallet_subcommand(
...
+                .map(|(sp_code, amount)| {
+                    let script = sp_code.get_placeholder_p2tr_spk();
+                    (script, Amount::from_sat(*amount))
+                })
+                .collect();
+
+            if let Some(recipients) = maybe_recipients {
+                if send_all {
+                    tx_builder.drain_wallet().drain_to(recipients[0].0.clone());
+                } else {
```
tvpeter: Is it possible to drain to an SP address? if so, it will be great to cover that as well.

src/handlers.rs
```diff
@@ -323,7 +333,160 @@ pub fn handle_offline_wallet_subcommand(
...
+                    derive_sp(&mut psbt, &k.xkey, &sp_recipients, &secp).expect("will fix later");
+                }
+                _ => unimplemented!("multi xkey signer"),
+            };
+
+            // Unfinalize PSBT to resign
+            for psbt_input in psbt.inputs.iter_mut() {
+                psbt_input.final_script_sig = None;
+                psbt_input.final_script_witness = None;
+            }
```
tvpeter, line 502: Won't it be ideal to check if they were actually finalized before resetting?

tvpeter (commented):
Thank you for working on this feature @nymius 

I left a few comments.

Pls remember to rebase too.

src/handlers.rs
```diff
@@ -323,7 +333,160 @@ pub fn handle_offline_wallet_subcommand(
...
+
+            let _resigned = wallet.sign(&mut psbt, SignOptions::default())?;
 
+            let raw_tx = psbt.extract_tx()?;
+            if cli_opts.pretty {
```
sdmg15, line 507: Here to actually make this "pretty", there's is this `shorten` function that can be used. I'm not though convinced if that's helpful for users wanting to copy the full raw tx maybe they should just not use the pretty format.

sdmg15 (commented):
tested ACK.

Script to use to reproduce

```bash
export NETWORK=regtest
export EXT_DESCRIPTOR='wpkh(tprv8ZgxMBicQKsPdMzWj9KHvoExKJDqfZFuT5D8o9XVZ3wfyUcnPNPJKncq5df8kpDWnMxoKbGrpS44VawHG17ZSwTkdhEtVRzSYXd14vDYXKw/0/*)'
export INT_DESCRIPTOR='wpkh(tprv8ZgxMBicQKsPdMzWj9KHvoExKJDqfZFuT5D8o9XVZ3wfyUcnPNPJKncq5df8kpDWnMxoKbGrpS44VawHG17ZSwTkdhEtVRzSYXd14vDYXKw/1/*)'
export DATABASE_TYPE="sqlite"
export CLIENT_TYPE="rpc"
export SERVER_URL="http://127.0.0.1:18443"
export COOKIE="/home/user/.bitcoin/regtest/.cookie"
export SP_RECIPIENT="sprt1qqdu4udd7wz7flf89m27teqzp8gxzvgpg3xudw89c0ce0f0pe0gwz6qkc4d7dg587w9rkaa0cj8rf3c28z0g45tgn0t7fzkk4s867w3eugga03ret"

raw_tx=$(cargo run -- wallet create_sp_tx --to-sp $SP_RECIPIENT:1000 | jq -r '.raw_tx' | tr -d '\n')
cargo run --features rpc -- wallet broadcast --tx $raw_tx
```

Image

src/handlers.rs
nymius: Good point, I'm not sure about the use cases of the `pretty` flag, but I'm against shortening the transaction.

src/handlers.rs
nymius: It shouldn't be necessary, as we are not passing a PSBT around, and the wallet always has all the keys required to sign, but is not difficult to implement. Above, `Wallet::sign` returns a boolean that indicates if the transaction was finalized or not, so I can check that.

src/handlers.rs
nymius: Yes, there are different cases for this though:
1. `CreateSpTx` receives only a `SilentPaymentCode` to drain funds.
2. `CreateSpTx` is called with the `--send_all` flag and multiple `SilentPaymentCode` or other addresses to drain funds.
3. `CreateSpTx` is called with the `--send_all` flag, no `SilentPaymentCode`, and a plain address.

I've implemented this contemplating 1 and 3 as successful cases, and any variant of 2 as an error.

sdmg15:
reACK https://github.com/bitcoindevkit/bdk-cli/commit/516af1b125250a6d74432a4768cf4d486b41eceb

src/commands.rs
```diff
@@ -182,6 +185,19 @@ pub enum CliSubCommand {
...
         shell: Shell,
     },
+    /// Silent payment code generation tool.
+    ///
+    /// Allows the encoding of two public keys into a silent payment code.
+    /// Useful to create silent payment transactions using fake silent payment codes.
+    #[cfg(feature = "silent-payments")]
+    SilentPaymentCode {
+        /// The scan public key to use on the silent payment code.
+        #[arg(long = "scan_public_key")]
```
tvpeter: nit: Since the documentation already stated that they are public keys, it will be easier for users to just call `scan_key`
```suggestion
        #[arg(long = "scan_key")]
```

src/commands.rs
```diff
@@ -182,6 +185,19 @@ pub enum CliSubCommand {
...
+        scan: bdk_sp::bitcoin::secp256k1::PublicKey,
+        /// The spend public key to use on the silent payment code.
+        #[arg(long = "spend_public_key")]
```
tvpeter:
```suggestion
        #[arg(long = "spend_key")]
```
Same suggestion as for `scan_key`

tvpeter (commented):
tACK 516af1b125250a6d74432a4768cf4d486b41eceb

Well done @nymius, I have left a few `nit`s for you.

Also, my final thoughts: 
i) With the current implementation, it is possible to send multiple outputs to the same SP code in the same tx. Can you consider checking for this?
i) Since `createtx` and `createsptx` are almost similar, it would be great to merge `createsptx` into the existing `createtx` to improve the UX for users, but allow the fn to detect an `SP` address and handle it internally. This can be in another PR.

nymius:
> Also, my final thoughts: i) With the current implementation, it is possible to send multiple outputs to the same SP code in the same tx. Can you consider checking for this?

I'm not sure I follow you here. The current implementation works with multiple SP codes or with a single one repeated multiple times. Do you mean if I have tested explicitly with `bdk-cli`? I can check it again.

> i) Since `createtx` and `createsptx` are almost similar, it would be great to merge `createsptx` into the existing `createtx` to improve the UX for users, but allow the fn to detect an `SP` address and handle it internally. This can be in another PR.

The idea of keeping them separated is that is easier to remove this command and replace it by a more mature silent payment sending command later. But we can work the details in another Issue/PR.

tvpeter:
> > Also, my final thoughts: i) With the current implementation, it is possible to send multiple outputs to the same SP code in the same tx. Can you consider checking for this?
> 
> I'm not sure I follow you here. The current implementation works with multiple SP codes or with a single one repeated multiple times. Do you mean if I have tested explicitly with `bdk-cli`? I can check it again.

The latter is what I am referring to; a single user can send more than one output to the same SP code e.g

```
... create_sp_tx  --to-sp tsp1q...mh:5000  --to-sp tsp1q...mh:3000

  ```
I don't know whether this is considered key reuse. That is why I called your attention to check if it is acceptable.

nymius:
> > > Also, my final thoughts: i) With the current implementation, it is possible to send multiple outputs to the same SP code in the same tx. Can you consider checking for this?
> > 
> > 
> > I'm not sure I follow you here. The current implementation works with multiple SP codes or with a single one repeated multiple times. Do you mean if I have tested explicitly with `bdk-cli`? I can check it again.
> 
> The latter is what I am referring to; a single user can send more than one output to the same SP code e.g
> 
> ```
> ... create_sp_tx  --to-sp tsp1q...mh:5000  --to-sp tsp1q...mh:3000
> ```
> 
> I don't know whether this is considered key reuse. That is why I called your attention to check if it is acceptable.

That's one of the features from silent payments, you can reuse the same silent payment code in the transaction and you're going to get two different output scripts. Have you checked the outputs of the transactions? The one corresponding to the 5000 amount did look similar to the one sending 3000 sats?

tvpeter:
> That's one of the features from silent payments, you can reuse the same silent payment code in the transaction and you're going to get two different output scripts. 

Alright then, that's fine.

> Have you checked the outputs of the transactions? The one corresponding to the 5000 amount did look similar to the one sending 3000 sats?

I did not run it now; I ran it during my last test. I will inspect the output after I test shortly.

tvpeter (approved):
tACK c57985475172fc610b3ce996338a43baef1c52d6

Thank you for adding this feature @nymius.

I have now confirmed that it indeed produced different outputs.

Image

tvpeter (approved):
tACK c57985475172fc610b3ce996338a43baef1c52d6

Thank you for adding this feature @nymius.

I have now confirmed that it indeed produced different outputs.

Image

PR #222, opened 2025-10-13
Update `bdk_kyoto` to `0.15.1`

rustaceanrob:
Checkpoints were removed from Kyoto to minimize trust in the library and reduce maintence burden. As such, there is no longer the ability to "skip blocks", at least on test networks. Mainnet offers two checkpoints, segwit and taproot activation, but these are included from genesis on test networks.

Broadcasting a transaction now awaits for the transaction to be gossiped, simplifying the logic here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,12 +259,12 @@ dependencies = [
 
 [[package]]
 name = "bdk_kyoto"
-version = "0.13.1"
+version = "0.15.1"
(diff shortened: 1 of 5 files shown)


tvpeter (approved):
tACK 70617f9e270fd194dbb977175811734a234a3d55

Thank you @rustaceanrob for this update.

I particularly like the improved logs as they are retained on the terminal.

Issue #218, opened 2025-09-08
NUMS_UNSPENDABLE_KEY_HEX is not supposed to be used as is (privacy concerns)


tvpeter:
I believe technically the `NUMS_UNSPENDABLE_KEY_HEX` is not supposed to be used as is (privacy concerns), though for now it's fine. Per [BIP-0431](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs), should add an issue to do the following:

"In order to avoid leaking the information that key path spending is not possible it is recommended to pick a fresh integer r in the range 0...n-1 uniformly at random and use H + rG as internal key. It is possible to prove that this internal key does not have a known discrete logarithm with respect to G by revealing r to a verifier who can then reconstruct how the internal key was created."

_Originally posted by @notmandatory in https://github.com/bitcoindevkit/bdk-cli/pull/208#discussion_r2330503782_

va-an:
If this is planned to be done not in 2.0.0, but in a subsequent version, then please assign this to me.

vadim-anfv:
@tvpeter @notmandatory ping - any thoughts on the question above? 
Happy to pick this up once there is a direction.

PR #225, opened 2025-11-19
feat(compile): compile taproot descriptor with randomized unspendable internal key

vadim-anfv:
### Description

Part of #218

Implements randomization of the unspendable internal key for taproot descriptors. This is the first part of #218, which consists of two parts:
1. **This PR**: Randomize unspendable internal key for taproot descriptor
2. **Follow-up**: Add verification command to ensure internal key is derived from NUMS key

Split into separate PRs for easier review and iteration, and to allow independent discussion of the verification command implementation, as one of the possible approaches could introduce breaking changes.

### Notes to the reviewers
The `compile` command now returns an additional `r` field for taproot descriptors (`-t tr`), containing the randomly generated internal key. Each compilation will produce a different internal key instead of using a fixed NUMS key.

Example output for taproot (first execution):
```
-> % bdk-cli compile "pk(A)" -t tr
{
  "descriptor": "tr(2dd09dd0355f4b2d5a4886de599786f3c0211652373221c87aba1cd1f7f1e593,pk(A))#anvu48aj",
  "r": "275a58827bd0ad459d6f92e083ddc3d99a03076155691680eb8f3b06380cdcfd"
}
```

Same descriptor compiled again produces different `r` and internal key:
```
-> % bdk-cli compile "pk(A)" -t tr
{
  "descriptor": "tr(801078f69dae7d95631723d4d13e6c32911633d227dcfc24c6b7e32e1e533e6c,pk(A))#f79rr82j",
  "r": "5e3ac63bb20d6a4bfff645279cc63a7472e18066da8826b13cbcb23aecb5c401"
}
```

Other descriptor types remain unchanged:
```
-> % bdk-cli compile "pk(A)" -t sh
{
  "descriptor": "sh(pk(A))#k80zhe7s"
}
```

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ tracing = "0.1.44"
 tracing-subscriber = "0.3.20"
 toml = "1.1.0"
 serde= {version = "1.0", features = ["derive"]}
+shlex = "1.3.0"
 
 # Optional dependencies
 bdk_bitcoind_rpc = { version = "0.21.0", features = ["std"], optional = true }
@@ -34,7 +35,6 @@ bdk_esplora = { version = "0.22.1", features = ["async-https", "tokio"], optiona
 bdk_kyoto = { version = "0.15.4", optional = true }
 bdk_redb = { version = "0.1.1", optional = true }
 bdk_sp = { version = "0.1.0", optional = true, git = "https://github.com/bitcoindevkit/bdk-sp", tag = "v0.1.0" }
-shlex = {  version = "1.3.0", optional = true }
 payjoin = { version = "=1.0.0-rc.1", features = ["v1", "v2", "io", "_test-utils"], optional = true}
 reqwest = { version = "0.13.2", default-features = false, optional = true }
 url = { version = "2.5.8", optional = true }
@@ -44,7 +44,7 @@ bdk_bip322 = { version = "0.1.0", optional = true }
 default = ["repl", "sqlite"]
 
 # To use the app in a REPL mode
-repl = ["shlex"]
+repl = []
 
 # Available database options
 sqlite = ["bdk_wallet/rusqlite"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,3 +69,6 @@ compiler = []
 
 # Experimental silent payment sending capabilities
 silent-payments = ["dep:bdk_sp"]
+
+[dev-dependencies]
+claims = "0.8.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ tracing-subscriber = "0.3.20"
 toml = "1.1.0"
 serde= {version = "1.0", features = ["derive"]}
 shlex = "1.3.0"
+tap = "1.0.1"
 
 # Optional dependencies
 bdk_bitcoind_rpc = { version = "0.21.0", features = ["std"], optional = true }
--- a/src/error.rs
+++ b/src/error.rs
@@ -49,6 +49,9 @@ pub enum BDKCliError {
     #[error("Miniscript error: {0}")]
     MiniscriptError(#[from] bdk_wallet::miniscript::Error),
 
+    #[error("Miniscript compiler error: {0}")]
+    MiniscriptCompilerError(#[from] bdk_wallet::miniscript::policy::compiler::CompilerError),
+
     #[error("ParseError: {0}")]
     ParseError(#[from] bdk_wallet::bitcoin::address::ParseError),
 
@@ -81,6 +84,10 @@ pub enum BDKCliError {
 
     #[error("Signer error: {0}")]
     SignerError(#[from] bdk_wallet::signer::SignerError),
+
+    #[cfg(feature = "compiler")]
+    #[error("Secp256k1 error: {0}")]
+    Secp256k1Error(#[from] bdk_wallet::bitcoin::secp256k1::Error),
 
     #[cfg(feature = "electrum")]
     #[error("Electrum error: {0}")]
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -40,12 +40,18 @@ use bdk_wallet::miniscript::miniscript;
 #[cfg(feature = "sqlite")]
 use bdk_wallet::rusqlite::Connection;
 use bdk_wallet::{KeychainKind, SignOptions, Wallet};
+
 #[cfg(feature = "compiler")]
 use bdk_wallet::{
-    bitcoin::XOnlyPublicKey,
+    bitcoin::{
+        XOnlyPublicKey,
+        key::{Parity, rand},
+        secp256k1::{Scalar, SecretKey},
+    },
     descriptor::{Descriptor, Legacy, Miniscript},
     miniscript::{Tap, descriptor::TapTree, policy::Concrete},
 };
+
 use clap::CommandFactory;
 use cli_table::{Cell, CellStruct, Style, Table, format::Justify};
 use serde_json::json;
@@ -1279,30 +1285,38 @@ pub(crate) fn handle_compile_subcommand(
     pretty: bool,
 ) -> Result<String, Error> {
     let policy = Concrete::<String>::from_str(policy.as_str())?;
-    let legacy_policy: Miniscript<String, Legacy> = policy
-        .compile()
-        .map_err(|e| Error::Generic(e.to_string()))?;
-    let segwit_policy: Miniscript<String, Segwitv0> = policy
-        .compile()
-        .map_err(|e| Error::Generic(e.to_string()))?;
-    let taproot_policy: Miniscript<String, Tap> = policy
-        .compile()
-        .map_err(|e| Error::Generic(e.to_string()))?;
+
+    let legacy_policy: Miniscript<String, Legacy> = policy.compile()?;
+    let segwit_policy: Miniscript<String, Segwitv0> = policy.compile()?;
+    let taproot_policy: Miniscript<String, Tap> = policy.compile()?;
+
+    let mut r = None;
 
     let descriptor = match script_type.as_str() {
         "sh" => Descriptor::new_sh(legacy_policy),
         "wsh" => Descriptor::new_wsh(segwit_policy),
         "sh-wsh" => Descriptor::new_sh_wsh(segwit_policy),
         "tr" => {
-            // For tr descriptors, we use a well-known unspendable key (NUMS point).
-            // This ensures the key path is effectively disabled and only script path can be used.
-            // See https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
+            // For tr descriptors, we use a randomized unspendable key (H + rG).
+            // This improves privacy by preventing observers from determining if key path spending is disabled.
+            // See BIP-341: https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
 
-            let xonly_public_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)
-                .map_err(|e| Error::Generic(format!("Invalid NUMS key: {e}")))?;
+            let secp = Secp256k1::new();
+            let r_secret = SecretKey::new(&mut rand::thread_rng());
+            r = Some(r_secret.display_secret().to_string());
+
+            let nums_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)?;
+            let nums_point = bdk_wallet::bitcoin::secp256k1::PublicKey::from_x_only_public_key(
+                nums_key,
+                Parity::Even,
+            );
+
+            let internal_key_point = nums_point.add_exp_tweak(&secp, &Scalar::from(r_secret))?;
+            let (xonly_internal_key, _) = internal_key_point.x_only_public_key();
 
             let tree = TapTree::Leaf(Arc::new(taproot_policy));
-            Descriptor::new_tr(xonly_public_key.to_string(), Some(tree))
+
+            Descriptor::new_tr(xonly_internal_key.to_string(), Some(tree))
         }
         _ => {
             return Err(Error::Generic(
@@ -1310,19 +1324,29 @@ pub(crate) fn handle_compile_subcommand(
             ));
         }
     }?;
+
     if pretty {
-        let table = vec![vec![
+        let mut rows = vec![vec![
             "Descriptor".cell().bold(true),
-            descriptor.to_string().cell(),
-        ]]
-        .table()
-        .display()
-        .map_err(|e| Error::Generic(e.to_string()))?;
+            shorten(&descriptor, 32, 29).cell(),
+        ]];
+
+        if let Some(r_value) = &r {
+            rows.push(vec!["r".cell().bold(true), shorten(r_value, 4, 4).cell()]);
+        }
+
+        let table = rows
+            .table()
+            .display()
+            .map_err(|e| Error::Generic(e.to_string()))?;
+
         Ok(format!("{table}"))
     } else {
-        Ok(serde_json::to_string_pretty(
-            &json!({"descriptor": descriptor.to_string()}),
-        )?)
+        let mut output = json!({"descriptor": descriptor});
+        if let Some(r_value) = r {
+            output["r"] = json!(r_value);
+        }
+        Ok(serde_json::to_string_pretty(&output)?)
     }
 }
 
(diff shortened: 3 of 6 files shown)


src/error.rs
```diff
@@ -45,6 +45,9 @@ pub enum BDKCliError {
     #[error("Miniscript error: {0}")]
     MiniscriptError(#[from] bdk_wallet::miniscript::Error),
 
+    #[error("Miniscript compiler error: {0}")]
+    MiniscriptCompilerError(#[from] bdk_wallet::miniscript::policy::compiler::CompilerError),
+
```
tvpeter, line 54: The above miniscript error wraps all the errors from miniscript including this one.

Cargo.toml
```diff
@@ -54,3 +54,6 @@ verify = []
 # Extra utility tools
 # Compile policies
 compiler = []
+
+[dev-dependencies]
+shlex = "1.3.0"
```
tvpeter: `shlex` is an optional dependency, so instead of re-installing it under dev dependencies, it's better to make it a non-optional dependency.

src/handlers.rs
```diff
@@ -997,50 +1003,63 @@ pub(crate) fn handle_compile_subcommand(
...
+            let secp = Secp256k1::new();
+            let (r_secret, r_point) = secp.generate_keypair(&mut rand::thread_rng());
+            r = Some(r_secret.display_secret().to_string());
+
+            let nums_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)?;
+            let nums_point = PublicKey::from_x_only_public_key(nums_key, Parity::Even);
 
-            let xonly_public_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)
-                .map_err(|e| Error::Generic(format!("Invalid NUMS key: {e}")))?;
+            let internal_key_point = nums_point.combine(&r_point)?;
```
tvpeter: Have you considered the `add_exp_tweak` method? I think it might be better suited for this use case because it knows one of the factors is a Generator point, whereas `combine` is for adding arbitrary public points.

tvpeter (changes requested):
Thanks for working on this @va-an 
I have left a few comments.

Cargo.toml
vadim-anfv: `shlex` needs to remain both optional and in dev-dependencies:
  - Optional: it's only used by the `repl` feature, so other builds don't need it
  - Dev-dependencies: CI runs `cargo test --no-default-features` which excludes `repl`, but tests still need `shlex` to compile

  Making it non-optional would unnecessarily add the dependency to all builds.

  Alternatively, we could accept the dependency overhead but simplify the code and CI in return?

src/error.rs
vadim-anfv: I added this error only for creating policies in a more concise way:
https://github.com/bitcoindevkit/bdk-cli/pull/225/files#diff-148e623ad4cdaab49b81901460099f4efe7900e8c57e4f7a0ff6778589be01f8L896-R905

Which way do you prefer more?
1) remove error, restore using err_map for policies
2) keep error and create policies without err_map

src/handlers.rs
vadim-anfv: Good point, thanks! Done.

src/handlers.rs
```diff
@@ -893,50 +899,62 @@ pub(crate) fn handle_compile_subcommand(
...
+            Descriptor::new_tr(xonly_internal_key.to_string(), Some(tree))
         }
         _ => {
             return Err(Error::Generic(
                 "Invalid script type. Supported types: sh, wsh, sh-wsh, tr".to_string(),
             ));
         }
-    }?;
+    }?
+    .to_string();
```
110CodingP: nit: ig we do not need this `to_string` here because of the `Display` implementation of `Descriptor<String>`.

110CodingP (commented):
Refactoring the tests to use`std::process::Command` instead of calling the internal function `handle_compile_subcommand` make the tests vulnerable to the state of environment variables and makes debugging harder in my opinion.  But I have no strong disagreement to the approach in the PR.

src/handlers.rs
vadim-anfv: Fixed, thanks!

110CodingP (approved):

va-an:
@110CodingP I decided to add the code for removing env vars in tests, as we discussed, so we don't forget to do this in future PRs. 

I verified locally that this works as expected — the command process will not receive these env vars from the parent process.
```rust
use std::{env, process::Command};

fn main() {
    unsafe {
        env::set_var("FOO", "1");
        env::set_var("BAR", "2");
        env::set_var("BAZ", "3");
    }

    let cmd = Command::new("printenv")
        .env_remove("FOO")
        .env_remove("BAR")
        .output()
        .unwrap();

    let stdout = String::from_utf8_lossy(&cmd.stdout);

    assert!(!stdout.contains("FOO"), "FOO should be removed");
    assert!(!stdout.contains("BAR"), "BAR should be removed");
    assert!(stdout.contains("BAZ"), "BAZ should still be present");

    println!("it's all good man");
}
```

Please take a look — https://github.com/bitcoindevkit/bdk-cli/pull/225/commits/afc76aa8d44eb609548d0923c59cd55f6785a6bd.

src/error.rs
tvpeter: I see your point. You can leave it as it is, though `map_err` is also okay.

Cargo.toml
tvpeter: While I understand your point, the `repl` feature that uses `shlex` is a default feature, and the package size is less than 20 KiB. Therefore, I would prefer it to be a non-optional dependency

src/handlers.rs
```diff
@@ -893,50 +899,61 @@ pub(crate) fn handle_compile_subcommand(
...
+
     if pretty {
-        let table = vec![vec![
-            "Descriptor".cell().bold(true),
-            descriptor.to_string().cell(),
-        ]]
-        .table()
-        .display()
-        .map_err(|e| Error::Generic(e.to_string()))?;
+        let mut rows = vec![vec!["Descriptor".cell().bold(true), descriptor.cell()]];
```
tvpeter: the `descriptor` value is quite long, can you apply the `[shorten](https://github.com/bitcoindevkit/bdk-cli/blob/fb7f6c60ca4ac1ad80750391435bbafd7346c714/src/utils.rs#L381)` fn so it displays better?

tvpeter (commented):
Thank you for updating the PR @va-an 

Once you address the little comments, I will do a final test.

tests/compile.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::process::Command;
+
+fn run_cmd(cmd: &str) -> Result<String, String> {
+    let full_cmd = format!("run --features compiler -- {}", cmd);
+    let args = shlex::split(&full_cmd).unwrap();
+
+    let output = Command::new("cargo")
+        .args(args)
+        .env_remove("NETWORK")
+        .env_remove("DATADIR")
```
110CodingP: nit: ig removing `DATADIR` env variable is not required? Since the compile command does not use it?

110CodingP (approved):

tests/compile.rs
vadim-anfv: You're right, `DATADIR` isn't used in compile command. 
I'm planning to reuse this command execution code in other tests, so I preemptively cleared all env vars.

Cargo.toml
vadim-anfv: Got it, done!

src/handlers.rs
vadim-anfv: Done! I also used `shorten()` for r.
I extracted shorten description into a variable to shorten only the internal key and leave the rest of the descriptor as is.

without pretty:
```bash
-> % cargo run --all-features -- compile "pk(ABC)" -t tr
{
  "descriptor": "tr(1985c2a86e01bbcdb1b5806422531a73c77e568ecfdd6d2863c158a76628ea50,pk(ABC))#6qsjh5q3",
  "r": "e38caa40d0db7cae7037336c982278a7fc6b770e7a80f8bbce0d79976a14d5e1"
}
```

with pretty:
```rust
-> % cargo run --all-features -- compile "pk(ABC)" -t tr --pretty
+------------+----------------------------------+
| Descriptor | tr(ccb4...b1d0,pk(ABC))#qrk923py |
+------------+----------------------------------+
| r          | a129...7a14                      |
+------------+----------------------------------+
```

src/handlers.rs
```diff
@@ -893,50 +899,70 @@ pub(crate) fn handle_compile_subcommand(
...
-        let table = vec![vec![
-            "Descriptor".cell().bold(true),
-            descriptor.to_string().cell(),
-        ]]
-        .table()
-        .display()
-        .map_err(|e| Error::Generic(e.to_string()))?;
+        let descriptor = shorten_descriptor.unwrap_or(descriptor);
+
+        let mut rows = vec![vec!["Descriptor".cell().bold(true), descriptor.cell()]];
```
tvpeter:
```suggestion
let mut rows = vec![vec!["Descriptor".cell().bold(true), shorten(descriptor, 32, 29).cell()]];
```

you can remove: 908,  931-934, 946

you can also revert 951; it should size most screens comfortably.

tvpeter (commented):
Hi @va-an, 

Could you please squash your commits into two or three commits and fix the shortening approach?

Aside from that, it looks good to go. Thank you.

src/handlers.rs
vadim-anfv: Ok, I’ll add these changes when I do the squash.

vadim-anfv: I made the changes, and only then remembered why I did it this way — if we apply shorten globally (not just for tr), we'll get an error for other descriptors because we're trying to shorten an already too short string.

Here's an example of how to reproduce this (if you apply your suggested changes locally):
```bash
-> % cargo run --all-features -- compile "pk(ABC)" -t tr --pretty
+------------+------------------------------------------------------------------+
| Descriptor | tr(cf07afc46391a3ce22c1024a682ca...fc9486809e6,pk(ABC))#3qlg4fwz |
+------------+------------------------------------------------------------------+
| r          | ee1c3591351b7c879fd415c0fbe3871bf61c752d7d35ed2ed94f0a278de9073d |
+------------+------------------------------------------------------------------+

-> % cargo run --all-features -- compile "pk(ABC)" -t wsh --pretty
thread 'main' (1258654) panicked at src/utils.rs:383:39:
byte index 32 is out of bounds of `wsh(pk(ABC))#8gmtcnaw`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Also, I'm not sure we should successfully compile a script like "pk(ABC)" at all, but we can discuss this in a separate issue.

tvpeter: Yeah, I noticed this bug and fixed it in #203 [here](https://github.com/bitcoindevkit/bdk-cli/pull/203/changes#diff-23af5356f6001cd3993cd3c801fb7716ea02d1e504081b9fb569332db6107e80R367-R370). We may have to wait for the PR to get merged.

tvpeter: > Also, I'm not sure we should successfully compile a script like "pk(ABC)" at all, but we can discuss this in a separate issue.

Well, policy compilation is actually from the upstream source but I have tested policies that failed. I will try to see if I can know what made this one to pass.

vadim-anfv: > > Also, I'm not sure we should successfully compile a script like "pk(ABC)" at all, but we can discuss this in a separate issue.
> 
> Well, policy compilation is actually from the upstream source but I have tested policies that failed. I will try to see if I can know what made this one to pass.

I just noticed that bdk-cli can create a descriptor that bitcoin-cli doesn't accept as valid:
```bash
-> % bdk-cli compile "pk(ABC)"
{
  "descriptor": "wsh(pk(ABC))#8gmtcnaw"
}

-> % bitcoin-cli deriveaddresses "wsh(pk(ABC))#8gmtcnaw"
error code: -5
error message:
pk(): key 'ABC' is not valid
```

This might confuse some users.

As a simple fix, we could explicitly state in the `compile` command documentation that we don't validate keys at compile time.

vadim-anfv: > Yeah, I noticed this bug and fixed it in #203 [here](https://github.com/bitcoindevkit/bdk-cli/pull/203/changes#diff-23af5356f6001cd3993cd3c801fb7716ea02d1e504081b9fb569332db6107e80R367-R370). We may have to wait for the PR to get merged.

Nice! Let's wait for #203 to merge then.
I've added "Depends on" to the PR description.
If we have a label like "blocked" or "do-not-merge", please add it to this PR.

tests/compile.rs
```diff
@@ -0,0 +1,78 @@
...
+//!
+//! Tests for compile command and subcommands
+
+use std::process::Command;
+
+fn run_cmd(cmd: &str) -> Result<String, String> {
+    let full_cmd = format!("run --features compiler -- {}", cmd);
+    let args = shlex::split(&full_cmd).unwrap();
+
+    let output = Command::new("cargo")
```
notmandatory: I don't see any reason to test the running of these commands. I know some older tests also do that but it's a pattern I don't want to continue and would like to remove. As far as I can tell these tests only verify the `clap` library is doing its job right, which we do not need to re-test.  

I'd rather see the tests only verify the handlers functions return what is expected.

notmandatory (commented):
Overall the code looks good, my only concern is with running shell commands as part of the testing, which I'd like to see removed.

Ultimately I'd like to figure out how to do the TR optimizatoin for the `pk(A)` policy so that it creates the descriptor `tr(A)` but I don't know if/how to do it or if something would need to be changed upstream.

src/handlers.rs
```diff
@@ -1014,50 +1022,61 @@ pub(crate) fn handle_compile_subcommand(
...
+
+            let nums_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)?;
+            let nums_point = PublicKey::from_x_only_public_key(nums_key, Parity::Even);
+
+            let internal_key_point = nums_point.add_exp_tweak(&secp, &Scalar::from(r_secret))?;
+            let (xonly_internal_key, _) = internal_key_point.x_only_public_key();
 
-            let xonly_public_key = XOnlyPublicKey::from_str(NUMS_UNSPENDABLE_KEY_HEX)
-                .map_err(|e| Error::Generic(format!("Invalid NUMS key: {e}")))?;
+            let tree = TapTree::Leaf(Arc::new(policy.compile::<Tap>()?));
```
vadim-anfv, line 1315: I'm planning to replace `compile()` with `compile_tr()` for the `tr` match branch, and moving the compile step inside the match is preparation for that.

See: https://github.com/bitcoindevkit/bdk-cli/issues/244

vadim-anfv: Nope, this won't happen - https://github.com/bitcoindevkit/bdk-cli/issues/244#issuecomment-4129988444.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 
 [dependencies]
-bdk_wallet = { version = "2.1.0", features = ["rusqlite", "keys-bip39", "compiler", "std"] }
+bdk_wallet = { version = "=2.1.0", features = ["rusqlite", "keys-bip39", "compiler", "std"] }
```
vadim-anfv: See #254

va-an:
Hey @tvpeter, is there anything else you'd like me to do on my end? Should I squash the follow-up commits down again (a few have piled up since the last squash)?

Thanks!

tvpeter:
> Hey @tvpeter, is there anything else you'd like me to do on my end? Should I squash the follow-up commits down again (a few have piled up since the last squash)?

Hi @va-an, apologies for the delay. I will give it a final check shortly.

Cargo.toml
tvpeter: Pls revert because of [this](https://github.com/bitcoindevkit/bdk-cli/issues/254#issuecomment-4161840037)

tvpeter (commented):
hi @va-an , I have reviewed again and left one minor nit below.

Also, kindly remove these [lines](https://github.com/bitcoindevkit/bdk-cli/blob/f80d92a549ea648b65595a445cf1c6c187b7b337/src/handlers.rs#L1739-L1745) because the gating does not apply to compiler feature tests, and there is no way to run the compile feature tests except `all-features` are compiled.

Please, once you have effected these changes, tag me to approve.

Thank you.

vadim-anfv:
> hi @va-an , I have reviewed again and left one minor nit below.
> 
> Also, kindly remove these [lines](https://github.com/bitcoindevkit/bdk-cli/blob/f80d92a549ea648b65595a445cf1c6c187b7b337/src/handlers.rs#L1739-L1745) because the gating does not apply to compiler feature tests, and there is no way to run the compile feature tests except `all-features` are compiled.
> 
> Please, once you have effected these changes, tag me to approve.
> 
> Thank you.

@tvpeter done, pls have a look.
Should I squash commits into one after your review before merge?

tvpeter (approved):
Thank you @va-an.

While it is not mandatory, you can squash 05a248cbd0bde6ecd448369067f71a5518bd743a, 
1ed41d21ec0c9e898e231b958a787f96dcf9fdba and bc8135d8d3871c2087b98187caab40cae58858d8.

Also, please revert the bdK_wallet version to 2.1, as some features in versions above 2.1 have not yet been implemented in the CLI.

Thank you

vadim-anfv:
> Thank you @va-an.
> 
> While it is not mandatory, you can squash [05a248c](https://github.com/bitcoindevkit/bdk-cli/commit/05a248cbd0bde6ecd448369067f71a5518bd743a), [1ed41d2](https://github.com/bitcoindevkit/bdk-cli/commit/1ed41d21ec0c9e898e231b958a787f96dcf9fdba) and [bc8135d](https://github.com/bitcoindevkit/bdk-cli/commit/bc8135d8d3871c2087b98187caab40cae58858d8).
> 
> Also, please revert the bdK_wallet version to 2.1, as some features in versions above 2.1 have not yet been implemented in the CLI.
> 
> Thank you

@tvpeter phew, did it:
- rebased on master, dropped the merge commit and both `bdk_wallet` pin/bump commits
- added one cleanup commit on top to sync `Cargo.lock` with the new `Cargo.toml` entries

`bdk_wallet` is back on 2.1.0 in both `Cargo.toml` and `Cargo.lock`.

vadim-anfv:
@tvpeter rebased on master after #220 merge, history cleaned up.
@notmandatory would appreciate your review.

PR #235, opened 2026-01-06
CI: Update `continous_integration` and `audit` workflows

### LINKED ISSUE #239: RUSTSEC-2026-0007: Integer overflow in `BytesMut::reserve`
| Details |  |
| --- | --- |
| Package | `bytes` |
| Version | `1.11.0` |
| URL | https://github.com/advisories/GHSA-434x-w66g-qw3r |
| Patched Versions | >=1.11.1 |
| Unaffected Versions | <1.2.1 |
| Aliases | [CVE-2026-25541](https://nvd.nist.gov/vuln/detail/CVE-2026-25541), [GHSA-434x-w66g-qw3r](https://github.com/advisories/GHSA-434x-w66g-qw3r) |

In the unique reclaim path of `BytesMut::reserve`, the condition
```rs
if v_capacity >= new_cap + offset
```
uses an unchecked addition. When `new_cap + offset` overflows `usize` in release builds, this condition may incorrectly pass, causing `self.cap` to be set to a value that exceeds the actual allocated capacity. Subsequent APIs such as `spare_capacity_mut()` then trust this corrupted `cap` value and may create out-of-bounds slices, leading to UB.

This behavior is observable in release builds (integer overflow wraps), whereas debug builds panic due to overflow checks.

## PoC

```rs
use bytes::*;

fn main() {
    let mut a = BytesMut::from(&b"hello world"[..]);
    let mut b = a.split_off(5);

    // Ensure b becomes the unique owner of the backing storage
    drop(a);

    // Trigger overflow in new_cap + offset inside reserve
    b.reserve(usize::MAX - 6);

    // This call relies on the corrupted cap and may cause UB & HBO
    b.put_u8(b'h');
}
```

# Workarounds

Users of `BytesMut::reserve` are only affected if integer overflow checks are configured to wrap. When integer overflow is configured to panic, this issue does not apply.

tvpeter:
### Description

This PR updates the continuous integration workflow and fixes some Clippy issues.

Also fixes #239 

### Notes to the reviewers

- update `actions/checkout` to v6
- add a separate job for Clippy 
- automate manual caching in the `build-test` job
- replace `actions-r/toolchain` with `actions-rust-
lang/setup-rust-toolchain` for toolchain setup
-fix clippy warnings
- add prepush command to Justfile
- Add `tool-version` parameter in audit workflow that forces the `actions-rust-lang/audit` to use a version that supports CVSS 4.0
- update transitive dependency `bytes` to v1.11.1

#### All Submissions:

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -27,11 +27,8 @@ test:
 
 # checks before pushing
 pre-push:
-    cargo build --features default
     cargo test --features default
-    cargo build --no-default-features
     cargo test --no-default-features
-    cargo build --all-features
     cargo test --all-features
     cargo clippy --no-default-features --all-targets -- -D warnings
     cargo clippy --all-features --all-targets -- -D warnings
--- a/src/payjoin/mod.rs
+++ b/src/payjoin/mod.rs
@@ -608,16 +608,12 @@ impl<'a> PayjoinManager<'a> {
                                         return Err(ImplementationError::from("Cannot find the transaction in the mempool or the blockchain"));
                                     };
 
-                                    let is_seen = match tx_details.chain_position {
-                                        bdk_wallet::chain::ChainPosition::Confirmed { .. } => true,
-                                        bdk_wallet::chain::ChainPosition::Unconfirmed { first_seen: Some(_), .. } => true,
-                                        _ => false
-                                    };
+                                    let is_seen = matches!(tx_details.chain_position, bdk_wallet::chain::ChainPosition::Confirmed { .. } | bdk_wallet::chain::ChainPosition::Unconfirmed { first_seen: Some(_), .. });
 
                                     if is_seen {
                                         return Ok(Some(tx_details.tx.as_ref().clone()));
                                     }
-                                    return Err(ImplementationError::from("Cannot find the transaction in the mempool or the blockchain"));
+                                Err(ImplementationError::from("Cannot find the transaction in the mempool or the blockchain"))
                                 },
                                 |outpoint| {
                                     let utxo = self.wallet.get_utxo(outpoint);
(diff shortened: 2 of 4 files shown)


Justfile
```diff
@@ -25,6 +25,18 @@ build: fmt
...
+pre-push:
+    cargo build --features default
+    cargo test --features default
+    cargo build --no-default-features
+    cargo test --no-default-features
+    cargo build --all-features
+    cargo test --all-features
+    cargo clippy --no-default-features --all-targets -- -D warnings
+    cargo clippy --all-features --all-targets -- -D warnings
+    cargo fmt --all -- --check
```
vadim-anfv, line 35: We can drop cargo build steps — cargo test already compiles the code implicitly before running.

```suggestion
pre-push:
    cargo test --features default
    cargo test --no-default-features
    cargo test --all-features
    cargo clippy --no-default-features --all-targets -- -D warnings
    cargo clippy --all-features --all-targets -- -D warnings
    cargo fmt --all -- --check
```

notmandatory: Agree that test will build so don't need the separate build lines. But test won't re-build if build just ran. So either way I think this is OK.

src/payjoin/mod.rs
```diff
@@ -609,16 +608,19 @@ impl<'a> PayjoinManager<'a> {
...
-                                    let is_seen = match tx_details.chain_position {
-                                        bdk_wallet::chain::ChainPosition::Confirmed { .. } => true,
-                                        bdk_wallet::chain::ChainPosition::Unconfirmed { first_seen: Some(_), .. } => true,
-                                        _ => false
-                                    };
+                                    // let is_seen = match tx_details.chain_position {
+                                    //     bdk_wallet::chain::ChainPosition::Confirmed { .. } => true,
+                                    //     bdk_wallet::chain::ChainPosition::Unconfirmed { first_seen: Some(_), .. } => true,
+                                    //     _ => false
+                                    // };
```
vadim-anfv: Should we keep it?

PR #236, opened 2026-01-07
feat: add BIP353 DNS payment instructions

sdmg15:
### Description

This PR adds support for BIP353 payment instructions. 
The following notable changes are done: 

- Adding of an option `resolve_dns_recipient` which receives a Human Readable Name (HRN) and returns the associated address
- Modify the `create_tx` for it to support user specifying recipients as `HRN`.

### Notes to the reviewers

Here is a list of HRN that could be used to test the implementation: 
- pay@dunxen.dev: Supports BOLT 12 Payment method only (CAUTION: mainnet)
- send.some@satsto.me: Supports BOLT 12 payment and On Chain (CAUTION: mainnet)
- Testing resolving: `bdk-cli -n testnet resolve_dns_recipient testnet@bitdevsyde.org`
- Testing create tx: `bdk-cli -n testnet wallet -w regtest_default_wallet create_tx --to "bcrt1qe497k549sgw9trzym50tdlegq3xx4apjqynjqm:1000" --to_dns "testnet@bitdevsyde.org:5000"`

## Changelog notice

- Add top level command `resolve_dns_recipient` to resolve the DNS payment instructions
- Add option `--to_dns` to `create_tx` command to allow sending to a Human Readable Name (HRN)

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ payjoin = { version = "0.25.0", features = ["v1", "v2", "io", "_test-utils"], op
 reqwest = { version = "0.13.2", default-features = false, features = ["rustls"], optional = true }
 url = { version = "2.5.8", optional = true }
 bdk_bip322 = { version = "0.1.0", optional = true }
+bitcoin-payment-instructions = { version = "0.7.0", optional = true}
 
 [features]
 default = ["repl", "sqlite"]
@@ -55,7 +56,8 @@ redb = ["bdk_redb"]
 cbf = ["bdk_kyoto", "_payjoin-dependencies"]
 electrum = ["bdk_electrum", "_payjoin-dependencies"]
 esplora = ["bdk_esplora", "_payjoin-dependencies"]
-rpc = ["bdk_bitcoind_rpc", "_payjoin-dependencies"] 
+rpc = ["bdk_bitcoind_rpc", "_payjoin-dependencies"]
+dns_payment = ["bitcoin-payment-instructions"] 
 
 # Internal features
 _payjoin-dependencies = ["payjoin", "reqwest", "url", "sqlite"]
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -380,7 +380,7 @@ pub enum OfflineWalletSubCommand {
         /// Adds a recipient to the transaction.
         // Clap Doesn't support complex vector parsing https://github.com/clap-rs/clap/issues/1704.
         // Address and amount parsing is done at run time in handler function.
-        #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
+        #[arg(env = "ADDRESS:SAT", long = "to", value_parser = parse_recipient)]
         recipients: Vec<(ScriptBuf, u64)>,
         #[cfg(feature = "dns_payment")]
         /// Adds DNS recipients to the transaction
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -389,7 +389,7 @@ pub enum OfflineWalletSubCommand {
         dns_recipients: Vec<(String, u64)>,
         #[cfg(feature = "dns_payment")]
         /// Custom resolver DNS IP to be used for resolution.
-        #[arg(long = "dns_resolver", default_value = "8.8.8.8:53")]
+        #[arg(long = "dns_resolver", default_value = "8.8.8.8")]
         dns_resolver: String,
         /// Sends all the funds (or all the selected utxos). Requires only one recipient with value 0.
         #[arg(long = "send_all", short = 'a')]
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -16,7 +16,6 @@
 #[cfg(feature = "silent-payments")]
 use {crate::utils::parse_sp_code_value_pairs, bdk_sp::encoding::SilentPaymentCode};
 
-
 use bdk_wallet::bitcoin::{
     Address, Network, OutPoint, ScriptBuf,
     bip32::{DerivationPath, Xpriv},
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -16,6 +16,7 @@
 #[cfg(feature = "silent-payments")]
 use {crate::utils::parse_sp_code_value_pairs, bdk_sp::encoding::SilentPaymentCode};
 
+
 use bdk_wallet::bitcoin::{
     Address, Network, OutPoint, ScriptBuf,
     bip32::{DerivationPath, Xpriv},
@@ -23,6 +24,8 @@ use bdk_wallet::bitcoin::{
 use clap::{Args, Parser, Subcommand, ValueEnum, value_parser};
 use clap_complete::Shell;
 
+#[cfg(feature = "dns_payment")]
+use crate::utils::parse_dns_recipient;
 #[cfg(any(feature = "electrum", feature = "esplora", feature = "rpc"))]
 use crate::utils::parse_proxy_auth;
 use crate::utils::{parse_address, parse_outpoint, parse_recipient};
@@ -198,6 +201,16 @@ pub enum CliSubCommand {
         #[arg(long = "spend_key")]
         spend: bdk_sp::bitcoin::secp256k1::PublicKey,
     },
+
+    #[cfg(feature = "dns_payment")]
+    /// Resolves BIP-353 DNS payment instructions for a human-readable name.
+    ResolveDnsRecipient {
+        /// Human-readable name (e.g. user@domain.com)
+        hrn: String,
+        /// DNS resolver address
+        #[arg(long, default_value = "8.8.8.8")]
+        resolver: String,
+    },
 }
 
 /// Wallet operation subcommands.
@@ -370,6 +383,14 @@ pub enum OfflineWalletSubCommand {
         // Address and amount parsing is done at run time in handler function.
         #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
         recipients: Vec<(ScriptBuf, u64)>,
+        #[cfg(feature = "dns_payment")]
+        /// Adds DNS recipients to the transaction
+        #[arg(long = "to_dns", value_parser = parse_dns_recipient)]
+        dns_recipients: Vec<(String, u64)>,
+        #[cfg(feature = "dns_payment")]
+        /// Custom resolver DNS IP to be used for resolution.
+        #[arg(long = "dns_resolver", default_value = "8.8.8.8:53")]
+        dns_resolver: String,
         /// Sends all the funds (or all the selected utxos). Requires only one recipient with value 0.
         #[arg(long = "send_all", short = 'a')]
         send_all: bool,
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -380,14 +380,57 @@ pub enum OfflineWalletSubCommand {
         /// Adds a recipient to the transaction.
         // Clap Doesn't support complex vector parsing https://github.com/clap-rs/clap/issues/1704.
         // Address and amount parsing is done at run time in handler function.
+        #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
+        recipients: Vec<(ScriptBuf, u64)>,
+        /// Sends all the funds (or all the selected utxos). Requires only one recipient with value 0.
+        #[arg(long = "send_all", short = 'a')]
+        send_all: bool,
+        /// Enables Replace-By-Fee (BIP125).
+        #[arg(long = "enable_rbf", short = 'r', default_value_t = true)]
+        enable_rbf: bool,
+        /// Make a PSBT that can be signed by offline signers and hardware wallets. Forces the addition of `non_witness_utxo` and more details to let the signer identify the change output.
+        #[arg(long = "offline_signer")]
+        offline_signer: bool,
+        /// Selects which utxos *must* be spent.
+        #[arg(env = "MUST_SPEND_TXID:VOUT", long = "utxos", value_parser = parse_outpoint)]
+        utxos: Option<Vec<OutPoint>>,
+        /// Marks a utxo as unspendable.
+        #[arg(env = "CANT_SPEND_TXID:VOUT", long = "unspendable", value_parser = parse_outpoint)]
+        unspendable: Option<Vec<OutPoint>>,
+        /// Fee rate to use in sat/vbyte.
+        #[arg(env = "SATS_VBYTE", short = 'f', long = "fee_rate")]
+        fee_rate: Option<f32>,
+        /// Selects which policy should be used to satisfy the external descriptor.
+        #[arg(env = "EXT_POLICY", long = "external_policy")]
+        external_policy: Option<String>,
+        /// Selects which policy should be used to satisfy the internal descriptor.
+        #[arg(env = "INT_POLICY", long = "internal_policy")]
+        internal_policy: Option<String>,
+        /// Optionally create an OP_RETURN output containing given String in utf8 encoding (max 80 bytes)
+        #[arg(
+            env = "ADD_STRING",
+            long = "add_string",
+            short = 's',
+            conflicts_with = "add_data"
+        )]
+        add_string: Option<String>,
+        /// Optionally create an OP_RETURN output containing given base64 encoded String. (max 80 bytes)
+        #[arg(
+            env = "ADD_DATA",
+            long = "add_data",
+            short = 'o',
+            conflicts_with = "add_string"
+        )]
+        add_data: Option<String>, //base 64 econding
+    },
+    #[cfg(feature = "dns_payment")]
+    /// Creates a new unsigned transaction from DNS payment instructions.
+    CreateDnsTx {
+        /// Adds a recipient to the transaction.
         #[arg(env = "ADDRESS:SAT", long = "to", value_parser = parse_recipient)]
         recipients: Vec<(ScriptBuf, u64)>,
-        #[cfg(feature = "dns_payment")]
-        /// Adds DNS recipients to the transaction
         #[arg(long = "to_dns", value_parser = parse_dns_recipient)]
         dns_recipients: Vec<(String, u64)>,
-        #[cfg(feature = "dns_payment")]
-        /// Custom resolver DNS IP to be used for resolution.
         #[arg(long = "dns_resolver", default_value = "8.8.8.8")]
         dns_resolver: String,
         /// Sends all the funds (or all the selected utxos). Requires only one recipient with value 0.
--- a/src/dns_payment_instructions.rs
+++ b/src/dns_payment_instructions.rs
@@ -1,0 +1,287 @@
+use bdk_wallet::bitcoin::{Address, Amount, Network};
+use bitcoin_payment_instructions::{
+    FixedAmountPaymentInstructions, ParseError, PaymentInstructions, PaymentMethod,
+    PaymentMethodType, amount, dns_resolver::DNSHrnResolver,
+};
+use cli_table::{Cell, Style, Table, format::Justify};
+use core::{net::SocketAddr, str::FromStr};
+
+use crate::{error::BDKCliError as Error, utils::shorten};
+
+#[derive(Debug)]
+pub struct Payment {
+    pub payment_methods: Vec<PaymentMethod>,
+    pub min_amount: Option<Amount>,
+    pub max_amount: Option<Amount>,
+    pub description: Option<String>,
+    pub expected_amount: Option<Amount>,
+    pub receiving_addr: Option<Address>,
+    pub hrn: String,
+    pub notes: String,
+}
+
+impl Payment {
+    pub fn display(&self, pretty: bool) -> Result<String, Error> {
+        let mut methods: Vec<String> = Vec::new();
+        self.payment_methods.iter().for_each(|pm| match pm {
+            bitcoin_payment_instructions::PaymentMethod::LightningBolt11(bolt11) => {
+                methods.push(format!("Bolt 11 invoice ({})", shorten(bolt11, 20, 15)))
+            }
+            bitcoin_payment_instructions::PaymentMethod::LightningBolt12(offer) => {
+                methods.push(format!("Bolt 12 invoice ({})", shorten(offer, 20, 15)))
+            }
+            bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
+                methods.push(format!("On chain ({})", address))
+            }
+            bitcoin_payment_instructions::PaymentMethod::Cashu(csh) => {
+                methods.push(format!("Cashu payment ({})", shorten(csh, 20, 15)))
+            }
+        });
+
+        if pretty {
+            let mut table = vec![vec![
+                "HRN".cell().bold(true),
+                self.hrn.to_string().cell().justify(Justify::Right),
+            ]];
+
+            if let Some(min_amnt) = self.min_amount {
+                table.push(vec![
+                    "Min amount".cell().bold(true),
+                    min_amnt.to_string().cell().justify(Justify::Right),
+                ]);
+            }
+
+            if let Some(max_amnt) = self.max_amount {
+                table.push(vec![
+                    "Max amount".cell().bold(true),
+                    max_amnt.to_string().cell().justify(Justify::Right),
+                ]);
+            }
+
+            if let Some(send_amnt) = self.expected_amount {
+                table.push(vec![
+                    "Expected Amount to send".cell().bold(true),
+                    send_amnt.to_string().cell().justify(Justify::Right),
+                ]);
+            }
+
+            if let Some(descr) = &self.description {
+                table.push(vec![
+                    "Description".cell().bold(true),
+                    descr.cell().justify(Justify::Right),
+                ]);
+            }
+
+            table.push(vec![
+                "Accepted methods".cell().bold(true),
+                methods.join(", ").cell().justify(Justify::Right),
+            ]);
+            table.push(vec![
+                "Notes".cell().bold(true),
+                self.notes.clone().cell().justify(Justify::Right),
+            ]);
+
+            let table = table
+                .table()
+                .display()
+                .map_err(|e| Error::Generic(e.to_string()))?;
+
+            Ok(format!("{table}"))
+        } else {
+            Ok(serde_json::to_string_pretty(&serde_json::json!({
+                    "hrn": self.hrn,
+                    "payment_methods": methods,
+                    "description": self.description,
+                    "min_amount": self.min_amount,
+                    "max_amount": self.max_amount,
+                    "expected_amount_to_send": self.expected_amount,
+                    "notes": self.notes
+            }))?)
+        }
+    }
+}
+pub(crate) async fn parse_dns_instructions(
+    hrn: &str,
+    network: Network,
+    resolver_ip: String,
+) -> Result<(DNSHrnResolver, PaymentInstructions), ParseError> {
+    let ip_address = if resolver_ip.contains(':') {
+        resolver_ip
+    } else {
+        format!("{resolver_ip}:53")
+    };
+
+    let sock_addr = SocketAddr::from_str(&ip_address).map_err(|_| {
+        ParseError::HrnResolutionError("Unable to create socket from provided address")
+    })?;
+    let resolver = DNSHrnResolver(sock_addr);
+    let instructions = PaymentInstructions::parse(hrn, network, &resolver, true).await?;
+    Ok((resolver, instructions))
+}
+
+fn get_onchain_info(
+    instructions: &FixedAmountPaymentInstructions,
+) -> Result<(Address, Amount), Error> {
+    // Look for on chain payment method as it's the only one we can support
+    let PaymentMethod::OnChain(addr) = instructions
+        .methods()
+        .iter()
+        .find(|ix| matches!(ix, PaymentMethod::OnChain(_)))
+        .ok_or(Error::Generic(
+            "Missing Onchain payment method option.".to_string(),
+        ))?
+    else {
+        return Err(Error::Generic("Unsupported payment method".to_string()));
+    };
+
+    let Some(onchain_amount) = instructions.onchain_payment_amount() else {
+        return Err(Error::Generic(
+            "On chain amount should be specified".to_string(),
+        ));
+    };
+
+    // We need this conversion since Amount from instructions is different from Amount from bitcoin
+    Ok((addr.clone(), Amount::from_sat(onchain_amount.milli_sats())))
+}
+
+pub async fn process_instructions(
+    amount_to_send: Amount,
+    payment_instructions: &PaymentInstructions,
+    resolver: DNSHrnResolver,
+) -> Result<Payment, Error> {
+    match payment_instructions {
+        PaymentInstructions::ConfigurableAmount(instructions) => {
+            // Look for on chain payment method as it's the only one we can support
+            if !instructions
+                .methods()
+                .any(|method| matches!(method.method_type(), PaymentMethodType::OnChain))
+            {
+                return Err(Error::Generic("Unsupported payment method".to_string()));
+            }
+
+            let min_amount = instructions
+                .min_amt()
+                .map(|amnt| Amount::from_sat(amnt.milli_sats()));
+
+            let max_amount = instructions
+                .max_amt()
+                .map(|amnt| Amount::from_sat(amnt.milli_sats()));
+
+            if min_amount.is_some_and(|min| amount_to_send < min) {
+                return Err(Error::Generic(
+                    format!(
+                        "Amount to send should be greater than min {}",
+                        min_amount.unwrap()
+                    )
+                    .to_string(),
+                ));
+            }
+
+            if max_amount.is_some_and(|max| amount_to_send > max) {
+                return Err(Error::Generic(
+                    format!(
+                        "Amount to send should be lower than max {}",
+                        max_amount.unwrap()
+                    )
+                    .to_string(),
+                ));
+            }
+
+            let fixed_instructions = instructions
+                .clone()
+                .set_amount(
+                    amount::Amount::from_sats(amount_to_send.to_sat()).unwrap(),
+                    &resolver,
+                )
+                .await
+                .map_err(|err| {
+                    Error::Generic(format!("Error occured while parsing instructions {err}"))
+                })?;
+
+            let onchain_details = get_onchain_info(&fixed_instructions)?;
+
+            Ok(Payment {
+                payment_methods: vec![PaymentMethod::OnChain(onchain_details.clone().0)],
+                min_amount,
+                max_amount,
+                description: instructions.recipient_description().map(|s| s.to_string()),
+                expected_amount: Some(onchain_details.1),
+                receiving_addr: Some(onchain_details.0.clone()),
+                hrn: instructions.human_readable_name().unwrap().to_string(),
+                notes: "".to_string(),
+            })
+        }
+
+        PaymentInstructions::FixedAmount(instructions) => {
+            let onchain_info = get_onchain_info(instructions)?;
+
+            Ok(Payment {
+                payment_methods: vec![PaymentMethod::OnChain(onchain_info.clone().0)],
+                min_amount: None,
+                max_amount: instructions
+                    .max_amount()
+                    .map(|amnt| Amount::from_sat(amnt.milli_sats())),
+                description: instructions.recipient_description().map(|s| s.to_string()),
+                expected_amount: Some(onchain_info.1),
+                receiving_addr: Some(onchain_info.0),
+                notes: "".to_string(),
+                hrn: instructions.human_readable_name().unwrap().to_string(),
+            })
+        }
+    }
+}
+
+/// Resolves the dns payment instructions found at the specified Human Readable Name
+pub async fn resolve_dns_recipient(
+    hrn: &str,
+    network: Network,
+    ip: String,
+) -> Result<Payment, ParseError> {
+    let (resolver, instructions) = parse_dns_instructions(hrn, network, ip).await?;
+
+    match instructions {
+        PaymentInstructions::ConfigurableAmount(ix) => {
+            let description = ix.recipient_description().map(|s| s.to_string());
+            let min_amount = ix.min_amt().map(|amnt| Amount::from_sat(amnt.milli_sats()));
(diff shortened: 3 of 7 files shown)


va-an:
@sdmg15 Hi! Could you make it compile?

Cargo.toml
```diff
@@ -36,6 +36,7 @@ shlex = {  version = "1.3.0", optional = true }
 payjoin = { version = "1.0.0-rc.1", features = ["v1", "v2", "io", "_test-utils"], optional = true}
 reqwest = { version = "0.12.23", default-features = false, optional = true }
 url = { version = "2.5.4", optional = true }
+bitcoin-payment-instructions = { version = "0.5.0", optional = true}
```
vadim-anfv: Let's use the newer version 0.7.0.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub struct Payment {
+    pub address: Address,
+    pub amount: Amount,
+    pub min_amount: Option<Amount>,
+    pub max_amount: Option<Amount>,
+    pub dnssec_proof: Option<Vec<u8>>,
+}
```
vadim-anfv, line 19:
```suggestion
#[derive(Debug)]
pub struct Payment {
    pub address: Address,
    pub amount: Amount,
    pub min_amount: Option<Amount>,
    pub max_amount: Option<Amount>,
}
```

Let's just remove unused fields instead of marking them as `dead_code`.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,107 @@
...
+fn process_fixed_instructions(
+    instructions: &FixedAmountPaymentInstructions,
+) -> Result<Payment, ParseError> {
+    // Look for on chain payment method as it's the only one we can support
+    let PaymentMethod::OnChain(addr) = instructions
+        .methods()
+        .iter()
+        .find(|ix| matches!(ix, PaymentMethod::OnChain(_)))
+        .map(|pm| pm)
+        .unwrap()
```
vadim-anfv:
```suggestion
    let PaymentMethod::OnChain(addr) = instructions
        .methods()
        .iter()
        .find(|ix| matches!(ix, PaymentMethod::OnChain(_)))
        .unwrap()
```

Let's remove the unnecessary `.map()` call.

src/commands.rs
```diff
@@ -291,13 +291,25 @@ pub enum OfflineWalletSubCommand {
...
     /// Returns the current wallet balance.
     Balance,
+    #[cfg(feature = "dns_payment")]
+    /// Resolves the given DNS payment instructions
+    ResolveDnsRecipient {
+        /// Human Readable Name to resolve
+        hrn: String,
+        /// The amount you're willing to send to the HRN
+        amount: u64,
+    },
```
vadim-anfv: `resolve_dns_recipient` only uses network from the wallet — it doesn't need descriptors, keys, or any wallet state. Having it under wallet forces the user to provide `--wallet` and have a saved config, which is unnecessary friction.

Consider moving it to a top-level CliSubCommand (alongside key, compile, descriptor) so it can be invoked as:
```bash
bdk-cli --network testnet resolve_dns_recipient matt@mattcorallo.com 1000
```

src/handlers.rs
```diff
@@ -331,8 +334,24 @@ pub fn handle_offline_wallet_subcommand(
...
+            let resolved = resolve_dns_recipient(&hrn, Amount::from_sat(amount), wallet.network())
+                .await
+                .map_err(|e| Error::Generic(format!("{:?}", e)))?;
+
+            Ok(serde_json::to_string_pretty(&json!({
+                    "hrn": hrn,
+                    "recipient": resolved.address,
+                    "min_amount": resolved.min_amount.unwrap_or_default(),
+                    "max_amount": resolved.max_amount.unwrap_or_default(),
+            }))?)
```
vadim-anfv: This handler always returns JSON regardless of `--pretty` flag. Other wallet subcommands (e.g. Balance, NewAddress) check cli_opts.pretty and render a table when it's set.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,107 @@
...
+            // Look for on chain payment method as it's the only one we can support
+            if instructions
+                .methods()
+                .find(|method| matches!(method.method_type(), PaymentMethodType::OnChain))
+                .is_none()
+            {
+                return Err(ParseError::InvalidInstructions(
+                    "Unsupported payment method",
+                ));
+            }
```
vadim-anfv, line 109: I think we should clearly separate the resolve and create_tx responsibilities. As a user, I'd expect `resolve_dns_recipient` to return all payment methods from the DNS record, even those that bdk-cli can't build a transaction for. 

Filtering to onchain-only makes sense in create_tx `--to_dns`, but not in a lookup command.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,107 @@
...
+
+            let min_amount = instructions
+                .min_amt()
+                .map(|amnt| Amount::from_sat(amnt.sats_rounding_up()));
+            let max_amount = instructions
+                .max_amt()
+                .map(|amnt| Amount::from_sat(amnt.sats_rounding_up()));
+
+            let fixed_instructions = instructions
+                .set_amount(
```
vadim-anfv, line 141: `amount` doesn't belong in `resolve_dns_recipient`. The resolve command should just show what's available for a given HRN - payment methods, min/max amounts, etc. `set_amount` is a payment preparation step (it may even contact the recipient to generate a BOLT12 invoice), not a lookup step.

I'd suggest removing the amount parameter from this command.

va-an (changes requested):
Thanks for working on this @sdmg15! 
I have left a few suggestions, please have a look.

src/commands.rs
sdmg15: Yes this is true, but not sure if we will really like to have `resolve_dns_recipient` solely as top level command.
I think instead of letting the user passing the `amount` we could just infer it from its current wallet balance and do the resolution with that. Doing it this way will make `resolve_dns_recipient` not look standalone.

src/dns_payment_instructions.rs
sdmg15: I think here that you're right. Please ignore the previous comment about the `amount` field becoming optional.

src/commands.rs
vadim-anfv: I see your point, but consider this use case: a user might want to resolve a BIP-353 address via bdk-cli but make the actual payment using a different wallet. This is especially relevant now since BIP-353 adoption is still limited - bdk-cli could serve as a resolution tool for users whose wallets don't support it yet. 

Forcing users who only need DNS resolution to create a wallet in bdk-cli first would be confusing and unnecessary.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub expected_amount: Option<Amount>,
+    pub receiving_addr: Option<Address>,
+    pub notes: String,
+}
+
+pub(crate) async fn parse_dns_instructions(
+    hrn: &str,
+    network: Network,
+) -> Result<PaymentInstructions, ParseError> {
+    let resolver = DNSHrnResolver(SocketAddr::from_str("8.8.8.8:53").expect("Should not fail."));
```
vadim-anfv: Let's add an optional argument for specifying a dns-resolver and set 8.8.8.8 as the default value:
```rust
#[arg(long = "dns-resolver", default_value = "8.8.8.8:53")]
dns_resolver: SocketAddr,
```

8.8.8.8 might be blocked in certain networks or countries, or the user might prefer their own DNS for privacy reasons.

src/handlers.rs
```diff
@@ -1060,6 +1081,94 @@ pub(crate) fn handle_compile_subcommand(
...
+            table.push(vec![
+                "Description".cell().bold(true),
+                descr.cell().justify(Justify::Right),
+            ]);
+        }
+
+        table.push(vec![
+            "Accepted methods".cell().bold(true),
+            methods.join(", ").cell().justify(Justify::Right),
+        ]);
```
vadim-anfv: Some strings (e.g. Bolt 12 offers) can be very long in `--pretty` output. I suggest applying the [`shorten`](https://github.com/bitcoindevkit/bdk-cli/blob/b9cf2acc5f10db46fa30777ff0910b8610a5fc33/src/utils.rs#L365) function to all payment method variants except `OnChain` in `--pretty` mode.

```suggestion
        for pm in &resolved.payment_methods {
            let val = match pm {
                bitcoin_payment_instructions::PaymentMethod::LightningBolt11(bolt11) => {
                    format!("Bolt 11 invoice ({})", shorten(bolt11, 20, 12))
                }
                bitcoin_payment_instructions::PaymentMethod::LightningBolt12(offer) => {
                    format!("Bolt 12 offer ({})", shorten(offer, 20, 12))
                }
                bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
                    format!("On chain ({})", address)
                }
                bitcoin_payment_instructions::PaymentMethod::Cashu(cashu) => {
                    format!("Cashu ({})", shorten(cashu, 20, 12))
                }
            };
            table.push(vec![
                "Method".cell().bold(true),
                val.cell().justify(Justify::Right),
            ]);
        }
```

src/handlers.rs
```diff
@@ -1060,6 +1081,94 @@ pub(crate) fn handle_compile_subcommand(
...
+        }
+        bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
+            methods.push(format!("On chain ({})", address.to_string()))
+        }
+        bitcoin_payment_instructions::PaymentMethod::Cashu(_) => {
+            methods.push("Cashu payment".to_string())
+        }
+    });
+
+    if pretty {
```
vadim-anfv: Consider moving the pretty output formatting into a separate function in `dns_payment_instructions.rs` to more explicitly separate logic from presentation.

src/handlers.rs
```diff
@@ -1060,6 +1081,94 @@ pub(crate) fn handle_compile_subcommand(
...
+    } else {
+        Ok(serde_json::to_string_pretty(&json!({
+                "hrn": hrn,
+                "payment_methods": methods,
+                "description": resolved.description,
+                "min_amount": resolved.min_amount,
+                "max_amount": resolved.max_amount,
+                "expected_amount_to_send": resolved.expected_amount,
+                "notes": resolved.notes
+        }))?)
```
vadim-anfv: Consider creating an object for command output like

```rust
pub struct ResolvedRecipient {
    pub hrn: String,
    pub methods: Vec<String>,
    pub description: Option<String>,
    pub min_amount: Option<Amount>,
    pub max_amount: Option<Amount>,
    pub expected_amount: Option<Amount>,
    pub notes: String,
}

impl ResolvedRecipient {
    pub fn render(&self, pretty: bool) -> Result<String, Error> { ... }
}
```

And then in the handler we can return the response something like:
```rust
Ok(resolved.render(cli_opts.pretty)?)
```

`struct Payment` already looks very similar, but we still need to perform 
transformations on it for display purposes. We could implement `render` for it 
instead of creating a new struct - the idea is to simplify the logic of 
working with this object.

Additionally, we might consider generalizing this approach by introducing a 
trait like `Renderer` or similar, which could allow for consistent output 
formatting across different command types.

src/handlers.rs
```diff
@@ -1060,6 +1081,94 @@ pub(crate) fn handle_compile_subcommand(
...
+        bitcoin_payment_instructions::PaymentMethod::LightningBolt12(offer) => {
+            methods.push(format!("Bolt 12 invoice ({})", offer.to_string()))
+        }
+        bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
+            methods.push(format!("On chain ({})", address.to_string()))
+        }
+        bitcoin_payment_instructions::PaymentMethod::Cashu(_) => {
+            methods.push("Cashu payment".to_string())
+        }
+    });
```
vadim-anfv: See https://github.com/bitcoindevkit/bdk-cli/pull/236/changes#r2805252489

va-an (commented):
@sdmg15 thanks for update and for the detailed description! 
Please take a look at the comments.

src/utils.rs
```diff
@@ -70,6 +67,17 @@ pub(crate) fn parse_recipient(s: &str) -> Result<(ScriptBuf, u64), String> {
     Ok((addr.script_pubkey(), val))
 }
 
+#[cfg(feature = "dns_payment")]
+/// Parse dns recipients in the form "test@me.com:10000" from cli input
+pub(crate) fn parse_dns_recipients(s: &str) -> Result<(String, u64), String> {
```
vadim-anfv: It would be better to rename it to `parse_dns_recipient`, because the function parses one recipient at a time.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub receiving_addr: Option<Address>,
+    pub hrn: String,
+    pub notes: String,
+}
+
+impl Payment {
+
+    pub fn display(&self, pretty: bool) -> Result<String, Error> {
+    let mut methods: Vec<String> = Vec::new();
+    self.payment_methods.iter().for_each(|pm| match pm {
```
vadim-anfv: It looks like you forgot to run `cargo fmt` 🙂 , please do so.

src/commands.rs
```diff
@@ -127,6 +130,10 @@ pub enum CliSubCommand {
     },
     /// List all saved wallet configurations.
     Wallets,
+
+    #[cfg(feature = "dns_payment")]
+    /// Resolves the given hrn payment instructions
+    ResolveDnsRecipient { hrn: String , resolver: Option<String>},
```
vadim-anfv:
```suggestion
    #[cfg(feature = "dns_payment")]
    /// Resolves BIP-353 DNS payment instructions for a human-readable name.
    ResolveDnsRecipient {
        /// Human-readable name (e.g. user@domain.com)
        hrn: String,
        /// DNS resolver address
        #[arg(long, default_value = "8.8.8.8")]
        resolver: String,
    },
```

This will make `--help` more useful:
```bash
-> % cargo run --all-features -- -n bitcoin resolve_dns_recipient --help
Resolves BIP-353 DNS payment instructions for a human-readable name

Usage: bdk-cli resolve_dns_recipient [OPTIONS] <HRN>

Arguments:
  <HRN>  Human-readable name (e.g. user@domain.com)

Options:
      --resolver <RESOLVER>  DNS resolver address [default: 8.8.8.8]
      --pretty               Output results in pretty format (instead of JSON)
  -h, --help                 Print help
```

Please note that the type for the resolver has changed.
See the suggested changes in the function itself.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,286 @@
...
+pub(crate) async fn parse_dns_instructions(
+    hrn: &str,
+    network: Network,
+    resolver_ip: Option<String>
+) -> Result<(DNSHrnResolver, PaymentInstructions), ParseError> {
+    
+    let mut ip_address = "8.8.8.8:53".to_string();
+    if let Some(res_addr) = resolver_ip {
+        ip_address = res_addr;
+    }
```
vadim-anfv:
```suggestion
    let ip_address = if resolver_ip.contains(':') {
        resolver_ip
    } else {
        format!("{resolver_ip}:53")
    };
```

Please note that `resolver_ip` should be passed to this function as a `String` (not `Option<String>`), because we added a default_value in `commands.rs`, as proposed [here](https://github.com/bitcoindevkit/bdk-cli/pull/236/changes#r2836453062).

Also, in this suggestion I slightly changed the semantics — the user can provide an IP without a port (port 53 will be used), or with a port (the user-specified port will be used). In the vast majority of cases, a DNS server uses the standard port, so we can simplify this a bit for the user.

The error message when creating `DNSHrnResolver` needs to be updated, since the user may provide an invalid value and receive a confusing "Should not fail" error.

va-an (commented):
@sdmg15 thanks for update!
I've added some suggestions, pls have a look.

va-an:
@sdmg15 thanks for updates!
Could you rebase?

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+            bitcoin_payment_instructions::PaymentMethod::LightningBolt12(offer) => {
+                methods.push(format!("Bolt 12 invoice ({})", shorten(offer, 20, 15)))
+            }
+            bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
+                methods.push(format!("On chain ({})", address))
+            }
+            bitcoin_payment_instructions::PaymentMethod::Cashu(csh) => {
+                methods.push(format!("Cashu payment ({})", shorten(csh, 20, 15)))
+            }
+        });
```
vadim-anfv: Consider `map + collect` here - drops the mutable accumulator and lets `match` return the `String` directly. Would need `PaymentMethod::*` in the `use` block.

```suggestion
        let methods: Vec<String> = self
            .payment_methods
            .iter()
            .map(|pm| match pm {
                LightningBolt11(bolt11) => {
                    format!("Bolt 11 invoice ({})", shorten(bolt11, 20, 15))
                }
                LightningBolt12(offer) => format!("Bolt 12 invoice ({})", shorten(offer, 20, 15)),
                OnChain(address) => format!("On chain ({})", address),
                Cashu(csh) => format!("Cashu payment ({})", shorten(csh, 20, 15)),
            })
            .collect();
```

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+            }
+            bitcoin_payment_instructions::PaymentMethod::LightningBolt12(offer) => {
+                methods.push(format!("Bolt 12 invoice ({})", shorten(offer, 20, 15)))
+            }
+            bitcoin_payment_instructions::PaymentMethod::OnChain(address) => {
+                methods.push(format!("On chain ({})", address))
+            }
+            bitcoin_payment_instructions::PaymentMethod::Cashu(csh) => {
+                methods.push(format!("Cashu payment ({})", shorten(csh, 20, 15)))
+            }
```
vadim-anfv: `shorten` runs before the `if pretty` check, so JSON output contains truncated Bolt 11 / Bolt 12 invoices and Cashu tokens - downstream tools can't use them.

Non-pretty output:
```rust
-> % cargo run --all-features -- -n bitcoin resolve_dns_recipient send.some@satsto.me
{
  "description": null,
  "expected_amount_to_send": null,
  "hrn": "send.some@satsto.me",
  "max_amount": null,
  "min_amount": null,
  "notes": "This is configurable payment instructions. You must send an amount between min_amount and max_amount if set.",
  "payment_methods": [
    "On chain (bc1qztwy6xen3zdtt7z0vrgapmjtfz8acjkfp5fp7l)",
    "Bolt 12 invoice (lno1zr5qyugqgskrk70k...ph4xrxtk2xc3lpq)"
  ]
}
```

The `shorten` calls must be moved into the `if pretty` branch only.

src/commands.rs
```diff
@@ -354,6 +367,14 @@ pub enum OfflineWalletSubCommand {
         // Address and amount parsing is done at run time in handler function.
         #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
         recipients: Vec<(ScriptBuf, u64)>,
+        #[cfg(feature = "dns_payment")]
+        /// Adds DNS recipients to the transaction
+        #[arg(long = "to_dns", value_parser = parse_dns_recipient)]
+        dns_recipients: Vec<(String, u64)>,
+        #[cfg(feature = "dns_payment")]
+        /// Custom resolver DNS IP to be used for resolution.
+        #[arg(long = "dns_resolver", default_value = "8.8.8.8:53")]
```
vadim-anfv: The default DNS resolver value is inconsistent between the two commands:
- `ResolveDnsRecipient::resolver` -> `"8.8.8.8"`
- `CreateTx::dns_resolver` -> `"8.8.8.8:53"`

`parse_dns_instructions` already appends `:53` when no port is present, so the `:53` suffix in `CreateTx` is redundant. Drop it to keep the defaults aligned.

Also worth aligning the flag names themselves - `--resolver` in one command and `--dns_resolver` in the other forces users to remember two names for the same thing.

Same value also keeps getting renamed as it travels through the call stack. For the resolve command:
- `commands.rs` -> `resolver`
- `handlers.rs::handle_resolve_dns_recipient_command(... resolver: String ...)`
- `dns_payment_instructions.rs::resolve_dns_recipient(... ip: String)`
- `dns_payment_instructions.rs::parse_dns_instructions(... resolver_ip: String)`

Three different names for the same value makes the flow harder to follow. Use `dns_resolver` everywhere - both as the CLI flag (`--dns_resolver`) and as the field/parameter name across all functions.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+                    "expected_amount_to_send": self.expected_amount,
+                    "notes": self.notes
+            }))?)
+        }
+    }
+}
+pub(crate) async fn parse_dns_instructions(
+    hrn: &str,
+    network: Network,
+    resolver_ip: String,
```
vadim-anfv: Take `&str` instead of `String` - `format!` already allocates, and it removes the need for `dns_resolver.clone()` at the call site.

src/handlers.rs
```diff
@@ -350,10 +357,30 @@ pub fn handle_offline_wallet_subcommand(
...
+                let mut recipients: Vec<(ScriptBuf, Amount)> = recipients
                     .into_iter()
                     .map(|(script, amount)| (script, Amount::from_sat(amount)))
                     .collect();
+
+                #[cfg(feature = "dns_payment")]
+                for recipient in dns_recipients {
+                    use crate::dns_payment_instructions::{
+                        parse_dns_instructions, process_instructions,
+                    };
```
vadim-anfv: Please move this `use` to the top of the file, next to the existing `dns_payment_instructions` import on line 16.

src/handlers.rs
```diff
@@ -350,10 +357,30 @@ pub fn handle_offline_wallet_subcommand(
...
+                    let (resolver, instructions) = parse_dns_instructions(
+                        &recipient.0,
+                        cli_opts.network,
+                        dns_resolver.clone(),
+                    )
+                    .await
+                    .map_err(|e| Error::Generic(format!("Parsing error occured {e:#?}")))?;
+                    let payment = process_instructions(amount, &instructions, resolver).await?;
+
+                    recipients.push((payment.receiving_addr.unwrap().into(), amount));
```
vadim-anfv: Two issues here:

1. `amount` is the user-supplied value, not `payment.expected_amount`. For `FixedAmount` instructions the recipient sets an exact amount, and the tx silently goes out with whatever the user typed instead.

2. `.unwrap()` relies on `process_instructions` always setting `receiving_addr: Some(...)` - an implicit invariant across two files.

Both resolved by the structural change in the top-level comment.

va-an (changes requested):
Thanks for addressing the previous round of feedback. Before tackling more changes - branch has merge conflicts with master, would you mind rebasing first? A few more comments below.

---
**Payment struct design**

`Payment` is shared between two flows with different needs.

In the send flow (`process_instructions` -> call site in `handlers.rs`) only `receiving_addr` is read, everything else is built and discarded. `.unwrap()` is needed only because the field is `Option<Address>` despite always being `Some` here.

In the resolve flow (`resolve_dns_recipient`) the same struct is filled with the opposite subset: `receiving_addr`/`expected_amount` always `None`, metadata fields populated for display. So every `Option<T>` here is "always Some" in one flow and "always None" in the other - the type system encodes nullability that doesn't actually exist in either path.

Suggested split:

```rust
// resolve flow - display only
pub struct ResolvedPaymentInfo {
    pub hrn: String,
    pub payment_methods: Vec<PaymentMethod>,
    pub description: Option<String>,
    pub min_amount: Option<Amount>,
    pub max_amount: Option<Amount>,
    pub notes: String,
}
// `display(pretty)` moves here

// send flow - return what's actually used
pub async fn process_instructions(...) -> Result<(Address, Amount), Error>
```

This resolves several inline comments: 
- unwrap on `receiving_addr`
- the `FixedAmount` amount mismatch
- both unwraps on `human_readable_name()`.

---

One small process note: please leave review threads open after addressing them - the convention in this repo is for the reviewer to resolve them once they confirm the fix.

vadim-anfv:
@sdmg15 when addressing the comments, could you split the changes into separate commits per independent fix instead of squashing everything into one? It makes the review much easier to follow (each commit maps to one review thread), and helps with bisect/revert later if needed. They can always be squashed at merge time.

va-an (changes requested):
I just found out that we're definitely dropping the `pretty` flag in #278, so please remove all the code related to this flag. This will also address some of my review comments.

Upd: although removing pretty output for this feature may create inconsistency with the current codebase, I expect we won't have any releases before #278 is merged, so this shouldn't become an issue for users.

@tvpeter please correct me if my assumption is wrong.

tvpeter:
> I just found out that we're definitely dropping the `pretty` flag in #278, so please remove all the code related to this flag. This will also address some of my review comments.
> 
> Upd: although removing pretty output for this feature may create inconsistency with the current codebase, I expect we won't have any releases before #278 is merged, so this shouldn't become an issue for users.
> 
> @tvpeter please correct me if my assumption is wrong.

Yes, that's correct. The `--pretty` flag is been removed. But you don't have to worry about removing it at your end. Once it is merged, I will clean it out.

src/handlers.rs
```diff
@@ -350,10 +357,30 @@ pub fn handle_offline_wallet_subcommand(
...
+                for recipient in dns_recipients {
+                    use crate::dns_payment_instructions::{
+                        parse_dns_instructions, process_instructions,
+                    };
+                    let amount = Amount::from_sat(recipient.1);
+                    let (resolver, instructions) = parse_dns_instructions(
+                        &recipient.0,
+                        cli_opts.network,
+                        dns_resolver.clone(),
+                    )
```
tvpeter: If it is possible, can you handle this DNS resolution concurrently? Else, print the status on the terminal so that users are aware that it is resolving the DNS.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+fn get_onchain_info(
+    instructions: &FixedAmountPaymentInstructions,
+) -> Result<(Address, Amount), Error> {
+    // Look for on chain payment method as it's the only one we can support
+    let PaymentMethod::OnChain(addr) = instructions
+        .methods()
+        .iter()
+        .find(|ix| matches!(ix, PaymentMethod::OnChain(_)))
```
tvpeter, line 78: I think you will have to reconsider prioritizing SP code for privacy concerns.

tvpeter (changes requested):
@sdmg15 thank you for working on this.

The comment I left regarding prioritizing silent payment code, I understand that this PR was worked on at a time when the silent payment feature was not yet merged, so you may have prioritized on-chain addresses. But since SP has been merged, and BIP353  (I believe) is designed to be paired with Silent Payments to prevent privacy leaks through permanent address reuse. So you might have to reconsider that implementation aspect.

Also, this feature has indicated using mainnet, and we have heavily discouraged users from using this app on mainnet. So you may also consider sharing another functional test network that users can use to play with. 

Thank you

src/dns_payment_instructions.rs
sdmg15: SP is still of type `OnChain` it's just the address derived that will change.

sdmg15:
>The comment I left regarding prioritizing silent payment code, I understand that this PR was worked on at a time when the silent payment feature was not yet merged, so you may have prioritized on-chain addresses. But since SP has been merged, and BIP353 (I believe) is designed to be paired with Silent Payments to prevent privacy leaks through permanent address reuse. So you might have to reconsider that implementation aspect.

Yes I think this will be great but the issue here is this will require some upstream changes to the `Address` or `PaymentMethod` type in order to be able to support the sp codes formats which I'm not sure they want to support experimental features/crates.


> Also, this feature has indicated using mainnet, and we have heavily discouraged users from using this app on mainnet. So you may also consider sharing another functional test network that users can use to play with.

I've updated the examples in the description with testnet DNS address

src/commands.rs
```diff
@@ -370,6 +382,14 @@ pub enum OfflineWalletSubCommand {
         // Address and amount parsing is done at run time in handler function.
         #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
         recipients: Vec<(ScriptBuf, u64)>,
+        #[cfg(feature = "dns_payment")]
+        /// Adds DNS recipients to the transaction
+        #[arg(long = "to_dns", value_parser = parse_dns_recipient)]
+        dns_recipients: Vec<(String, u64)>,
```
vadim-anfv: `--to` is `required = true`, so you can't create a tx with only a DNS recipient - even when paying solely via `--to_dns`, a `--to <address>` has to be supplied too.

```sh
$ cargo run --all-features -- \
    --network bitcoin \
    wallet -w dns_demo \
    create_tx --to_dns send.some@satsto.me:1000 \
    --fee_rate 1
error: the following required arguments were not provided:
  --to <RECIPIENTS>
```

At least one recipient should be required (`--to` or `--to_dns`), not `--to` unconditionally.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,202 @@
...
+        dns_resolver
+    } else {
+        &format!("{dns_resolver}:53")
+    };
+
+    let sock_addr = SocketAddr::from_str(ip_address).map_err(|_| {
+        ParseError::HrnResolutionError("Unable to create socket from provided address")
+    })?;
+    let resolver = DNSHrnResolver(sock_addr);
+    let instructions = PaymentInstructions::parse(hrn, network, &resolver, true).await?;
```
vadim-anfv: The DNS resolution has no timeout (neither here nor in `dnssec-prover`). With a user-supplied resolver, an unreachable one hangs the CLI instead of failing.

```sh
$ timeout 15 cargo run --all-features -- \
    --network bitcoin \
    resolve_dns_recipient send.some@satsto.me \
    --resolver 192.0.2.1
$ echo $?
124 # killed by the external timeout
```

A `tokio::time::timeout(...)` wrapper would fix it.

src/handlers.rs
```diff
@@ -542,10 +550,25 @@ pub fn handle_offline_wallet_subcommand(
...
-                let recipients = recipients
+                #[allow(unused_mut)]
+                let mut recipients: Vec<_> = recipients
                     .into_iter()
                     .map(|(script, amount)| (script, Amount::from_sat(amount)))
                     .collect();
+
+                #[cfg(feature = "dns_payment")]
+                for recipient in dns_recipients {
+                    println!("Resolving DNS instructions for recipient {}", recipient.0);
```
vadim-anfv: `println!` writes to stdout, mixing this progress line into the machine-readable command json output and breaking downstream parsing. Better as `log::info!` - clean output by default, and the user can opt into progress via `RUST_LOG`.

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl ResolvedPaymentInfo {
+    pub fn display(&self) -> Result<String, Error> {
+        let methods: Vec<String> = self
+            .payment_methods
+            .iter()
+            .map(|pm| match pm {
+                PaymentMethod::LightningBolt11(bolt11) => {
+                    format!("Bolt 11 invoice ({})", shorten(bolt11, 20, 15))
```
vadim-anfv: The Bolt11 invoice is truncated here, so the emitted value is broken - a clipped Bolt11 string isn't a valid invoice and can't be paid. The output should carry the full invoice, as Bolt12/OnChain/Cashu already do.

vadim-anfv (changes requested):
@sdmg15 thanks for the work on this, and for applying the previous round - the `Payment` split into `ResolvedPaymentInfo` and `process_instructions -> (Address, Amount)` came out clean.

Left a few inline comments below, plus one bigger-picture suggestion: unifying the recipients before the send logic. Right now `--to` and `--to_dns` are handled in separate places - DNS recipients are only resolved inside the non-`send_all` branch, so the building logic sees two different sources at two different points. If instead all `--to_dns` entries were resolved to `(address, amount)` and merged into the single `recipients` list up front - before any tx-building or `send_all` branching - the send path would just work off one unified list and wouldn't need to know where a recipient came from.

This would naturally fix a couple of the inline issues:
- `--to` no longer has to be `required` (validate "at least one recipient" on the merged list instead),
- `--send_all` would stop silently dropping `--to_dns` recipients (it currently only looks at `recipients[0]`).

Separating recipient parsing/resolution from the payment logic keeps the tx-building code simpler and removes these edge cases by construction.

Thanks again!

vadim-anfv (changes requested):
tACK 2ca719b

Tested `create_tx --to_dns` end-to-end on regtest (with a local patch to get past the network check). Works as expected - requesting 5000 sats and 250000 sats each produced an output of exactly that amount.

Right now `just pre-push` does not pass:

```
error: variable does not need to be mutable
   --> src/handlers.rs:532:13
    |
532 |             mut recipients,
    |             ----^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `-D unused-mut` implied by `-D warnings`
```

Other than that everything looks great.

sdmg15:
@va-an Thanks for the reviews. I've addressed that. We should be good to go.

vadim-anfv (changes requested):
Thanks a lot for all the work on this @sdmg15, great job - this has come together really nicely!

Just one small thing left: `Cargo.lock` contains changes unrelated to this PR - a stray `cargo update` bumped a lot of unrelated crates, including `bdk_wallet 2.1.0 -> 2.4.0`. Could you regenerate the lock from master so only the new dependency's entries land?

```bash
git checkout master -- Cargo.lock
cargo build --all-features
```

The lock also bumps `bdk_wallet` to 2.4.0 - this might not pass review from @tvpeter, see https://github.com/bitcoindevkit/bdk-cli/pull/225#pullrequestreview-4247884665.

Thanks again!

src/dns_payment_instructions.rs
```diff
@@ -0,0 +1,208 @@
...
+            if !instructions
+                .methods()
+                .any(|method| matches!(method.method_type(), PaymentMethodType::OnChain))
+            {
+                return Err(Error::Generic("Unsupported payment method".to_string()));
+            }
+
+            let min_amount = instructions
+                .min_amt()
+                .map(|amnt| Amount::from_sat(amnt.milli_sats()));
```
tvpeter: `Amount::from_sat` expects that the supplied amount be denominated in sats but you supplied `milli_sats`.

src/commands.rs
```diff
@@ -368,8 +380,16 @@ pub enum OfflineWalletSubCommand {
         /// Adds a recipient to the transaction.
         // Clap Doesn't support complex vector parsing https://github.com/clap-rs/clap/issues/1704.
         // Address and amount parsing is done at run time in handler function.
-        #[arg(env = "ADDRESS:SAT", long = "to", required = true, value_parser = parse_recipient)]
```
tvpeter, line 371: Making this optional here removes the validation that clap offers

src/handlers.rs
```diff
@@ -533,7 +537,14 @@ pub fn handle_offline_wallet_subcommand(
...
         }
         CreateTx {
+            #[cfg(feature = "dns_payment")]
+            mut recipients,
+            #[cfg(not(feature = "dns_payment"))]
             recipients,
+            #[cfg(feature = "dns_payment")]
+            dns_recipients,
+            #[cfg(feature = "dns_payment")]
+            dns_resolver,
```
tvpeter: I feel this introduces too many variables to combine with the existing CreateTx, so it will be better to add a new wallet command that handles DNS payment separately.

tvpeter (changes requested):
Following up on on my previous review comment about prioritizing SP and your response about the library, I have checked the library and confirmed that it does not include SP payment method. This is understandable as the author is lightning first and may not consider adding SP payment method.  So I think it is either you consider using another library that supports SP or provides mappings to non-reusable addresses or you handling that with the available toolset. But I don't think this PR as it is fully accomplishes the purpose of BIP353.

sdmg15:
> Following up on on my previous review comment about prioritizing SP and your response about the library, I have checked the library and confirmed that it does not include SP payment method. This is understandable as the author is lightning first and may not consider adding SP payment method.

The library isn't "lightning first". You can check this PR https://github.com/rust-bitcoin/bitcoin-payment-instructions/pull/19, there has been previous effort into adding that to the library but given that some upstream components aren't available they decided to postpone it.

>  But I don't think this PR as it is fully accomplishes the purpose of BIP353.

I think you're confusing the purpose of BIP353 and the one of BIP352. BIP353 for dns goal has never been about avoiding address reuse.  BIP352 just came and gave more sense to that proposal as combined together you have Human Readable Name that resolves to different addresses anytime when requested.

src/commands.rs
sdmg15: The validation is still done but at the command implementation level

tvpeter:
> > Following up on on my previous review comment about prioritizing SP and your response about the library, I have checked the library and confirmed that it does not include SP payment method. This is understandable as the author is lightning first and may not consider adding SP payment method.
> 
> The library isn't "lightning first". You can check this PR [rust-bitcoin/bitcoin-payment-instructions#19](https://github.com/rust-bitcoin/bitcoin-payment-instructions/pull/19), there has been previous effort into adding that to the library but given that some upstream components aren't available they decided to postpone it.
> 
> > But I don't think this PR as it is fully accomplishes the purpose of BIP353.
> 
> I think you're confusing the purpose of BIP353 and the one of BIP352. BIP353 for dns goal has never been about avoiding address reuse. BIP352 just came and gave more sense to that proposal as combined together you have Human Readable Name that resolves to different addresses anytime when requested.

So the whole point of my comment is that, if you can find a way around address reuse, it will serve the purpose of the BIP better.  See the BIP section on it [here](https://bips.xyz/353#address-reuse).

Thank you.

sdmg15:
Yeah I get the point.
I think that section is more a caution targeting users setting their DNS payment wallet. 

It's not meant for the application/resolver as you don't have control over that. Only the user will responsibly go and set a different wallet address in the DNS records. 

That's why BIP353 combined now with BIP352 effectively solves that as the user won't need to update the dns record any more but will get a new fresh Bitcoin address at any request.

src/handlers.rs
sdmg15: I don't know your final take on this but I tried doing it by adding `CreateDNSTx` or something similar which happens to just be a duplicate of `CreateTx` and introduce more codes. So let me know what you think should be best approach but I think they way it's done right isn't that much harmful

tvpeter (approved):
utACK d2790975aff929e6fd163b0db8af2ffa400bc712

PR #237, opened 2026-02-07
feat: add completions subcommand

### LINKED ISSUE #234: Add shell completions command
**Describe the enhancement**
Add shell completions support via a new `completions` subcommand, similar to how `rustup completions` works.
The command would generate shell completion scripts for bash, zsh, fish, powershell, and elvish using [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/).

Example usage:
```bash
bdk-cli completions bash > ~/.local/share/bash-completion/completions/bdk-cli
bdk-cli completions zsh > ~/.zfunc/_bdk-cli
```

**Use case**
bdk-cli has many subcommands and options. Typing out full command names gets tedious, so tab completion would make the CLI much nicer to use.

**Additional context**
I've already started looking into the implementation and can open a PR if this sounds good.

va-an:
### Description

Resolves #234.

Added `completions` subcommand to generate shell completions.

### How to test

Install bdk-cli:
```bash
cargo install --path . --all-features
```

See how to install completions for your shell:
```bash
bdk-cli completions --help
```

For example, to install completions for zsh:
```bash
bdk-cli completions zsh > ~/.zfunc/_bdk-cli
```

After restarting the shell, completions should work with Tab.

#### All Submissions:

#### New Features:

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,8 @@ This is available for wallet, key, repl and compile features. When ommitted, out
 bdk-cli completions --help
 ```
 
+Completion files are generated from the currently installed binary. After upgrading `bdk-cli`, re-run the `completions` command to keep them in sync with the new version.
+
 ## Saving and using wallet configurations
 
 The `wallet config` sub-command allows you to save wallet settings to a `config.toml` file in the default directory (`~/.bdk-bitcoin/`) or custom directory specified with the `--datadir` flag. This eliminate the need to repeatedly specify descriptors, client types, and other parameters for each command. Once configured, you can use any wallet command by simply specifying the wallet name. All other parameters are automatically loaded from the saved configuration.


src/commands.rs
```diff
@@ -127,7 +128,39 @@ pub enum CliSubCommand {
...
+    ///   #   autoload -Uz compinit && compinit
+    ///
+    /// Fish:
+    ///   bdk-cli completions fish > ~/.config/fish/completions/bdk-cli.fish
+    ///
+    /// PowerShell:
+    ///   bdk-cli completions powershell >> $PROFILE
+    ///
+    /// Elvish:
+    ///   bdk-cli completions elvish >> ~/.elvish/rc.elv
```
AmosOO7: I think this part should be moved to the `README.md` file.

vadim-anfv: This is for displaying the `bdk-cli completions --help` command. I made this by analogy with `rustup completions --help`.

I suggest keeping the detailed instructions for setting up completions in `bdk-cli completions --help`, since this relates only to a specific command, not to the entire program as a whole. 

But I think it would be useful to mention the new completions command in the README, as this will significantly simplify users' interaction with bdk-cli.

vadim-anfv: Added new block about completions to README.

va-an:
> Also I had to create the folders `bash-completion`, and `completions` in `~/.local/share` I don't know if this is peculiar to my case, but maybe it is something you can add in the readme file also to `mkdir` this folders so the command `bdk-cli completions bash > ~/.local/share/bash-completion/completions/bdk-cli` runs smoothly and for any other commands that might need it.

Thank you for catching this! 
I updated the help for completions, made the formatting cleaner, and added information about creating the directory.

src/commands.rs
AmosOO7: > This is for displaying the `bdk-cli completions --help` command. I made this by analogy with `rustup completions --help`.
> 
> I suggest keeping the detailed instructions for setting up completions in `bdk-cli completions --help`, since this relates only to a specific command, not to the entire program as a whole.
> 
> But I think it would be useful to mention the new completions command in the README, as this will significantly simplify users' interaction with bdk-cli.

Yeah, I agree to all you have said. I didn't think about it that was. Just wanted there to be a knowing about the `completion` in the `README.md` file, which I see you have done.

> Added new block about completions to README.

That's great

PR #230, opened 2025-12-21
Refactor sync_kyoto_client 

Mshehu5:
### Description

This PR addresses issues encountered while implementing persistence for Payjoin specifically around the BlockchainClient only being an owned variable rather than being able to be borrowed/referenced as &Blockchainclient.
While working on persistence I ran into problems while working on resume command which needs a blockchain client to resume states such as monitor_payjoin_proposal (receiver) and process_payjoin_proposal (sender)
Because BlockchainClient can only be owned the current design will require a function signature of passing two separate clients to resume sender and receiver states. I initially considered splitting the command into resume_send and resume_receive but this does not fully solve the issue. In particular the sender’s process_payjoin_proposal may call broadcast_transaction and potentially broadcast multiple transactions for persisted send entries stored in the database which still requires reusable access to the client.

This Ownership issue was previously mentioned in #200 and is also noted in a comment at the top of monitor_payjoin_proposal. It prevents the function from being able to resync multiple times and reliably detect when a transaction appears in the mempool. The root cause is that the Kyoto client Box<LightClient> is destructured and spawned into other tasks when passed through sync_kyoto_client making it unusable afterward.

What this PR changes
This PR fixes the issue by refactoring sync_kyoto_client 

- The logic responsible for running the Kyoto node and logger is moved into new_blockchain_client. This makes it that node is started at start of command and not during every sync

- Instead of returning a Box<lightClient> the function now returns a KyotoClientHandle. Previously the boxed client takes ownership when destructured inside sync_kyoto_client, preventing reuse/reference. 
With the new design sync_kyoto_client takes &KyotoClientHandle, allowing the client to be Refrenced which can be used  for syncing and broadcasting transactions without being owned

- Additionally monitor_payjoin_proposal is refactored to support resyncing demonstrating that the Kyoto client refactor successfully resolves the original limitations

### Notes to the reviewers

After refactor I tested the kyoto client on regtest (Cause I do not have access to a signet) I had to set a trusted peer in the code to connect with a cbf count of 1 this worked and I also made transaction using the steps below:
N.B Payjoin was also tested for the monitor_payjoin_proposal refactor using steps in project readme 
```
bitcoin.conf 

regtest=1
server=1
rpcuser=user
rpcpassword=password
rpcallowip=127.0.0.1
blockfilterindex=1
listen=1
fallbackfee=0.001

[regtest]
bind=127.0.0.1
port=18444
peerblockfilters=1
```

Step 1: Create transaction
```
[...]

Issue #149, opened 2020-07-27
Add Payjoin support


notmandatory:
**Description**

Implement (optional) [BIP-0078] [PayJoin] support in the `bdk-cli` wallet using the [rust bip78] project. PayJoin transactions provides a way to collaboratively create a transaction with a receiver in a way that prevents on-chain surveillance from using the [common input ownership heuristic]. Adding this feature to the `bdk-cli` project will demonstrate to other wallet builders using `bdk` how to add PayJoin support. 

**Expected Outcomes**  
 
* Integrate [BIP-0078] functionality as an optional feature in `bdk-cli` using the [rust bip78] crate
* Contribute upstream to the [rust bip78] project if any bugs or other issues are found
* Document how to use new PayJoin functionality with `bdk-cli` and common services such as BTCPayServer

**Resources**  

* [bdk](https://github.com/bitcoindevkit/bdk) and [bdk-cli] projects
* [rust-payjoin] project and [PayJoin]
* [The Rust Book](https://doc.rust-lang.org/stable/book/)
* [PayJoin adoption](https://en.bitcoin.it/wiki/PayJoin_adoption)

**Skills Required**  

* Rust language development, or similar language with willingness to learn rust
* Be able to send and receive bitcoin transactions on testnet with `bdk-cli` and other wallets
* Basic understanding bitcoin transaction building 

**Mentor(s)**  
@notmandatory

**Difficulty**  
Hard

**Competency Test**  

* [Install rust](https://www.rust-lang.org/tools/install)
* Clone and run [bdk-cli], create single signature descriptor wallet, receive and send testnet bitcoin
* Demonstrate rust coding skill:
  *  via simple rust program
  * OR by completing [rustlings] exercises: variables, functions, if, primitive types

[PayJoin]: https://payjoin.org/
[rust-payjoin]: https://github.com/payjoin/rust-payjoin
[rust bip78]: https://crates.io/crates/bip78
[BIP-0078]: https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki
[common input ownership heuristic]: https://en.bitcoin.it/wiki/Common-input-ownership_heuristic
[bdk-cli]: https://github.com/bitcoindevkit/bdk-cli
[rustlings]: https://github.com/rust-lang/rustlings

DanGould:
[...]

PR #242, opened 2026-02-16
Payjoin persistence

Mshehu5:
### Description
https://github.com/bitcoindevkit/bdk-cli/pull/230 needed to be merged for this to go through
Address https://github.com/bitcoindevkit/bdk-cli/issues/149 also follow up to #200 
This PR adds persistance to existing async payjoin integration 
This introduces neccessary database model and tables also  add commad for resume to allow interrupted sessions to be continued also a particular session either send or receive.
A history commad to view payjoin history and status has been added
  
### Notes to the reviewers

Step to review this include making a payjoin transaction 

Run a receiver to get a BIP21 URI then pass it to the sender as seen in docs https://github.com/bitcoindevkit/bdk-cli/blob/b9cf2acc5f10db46fa30777ff0910b8610a5fc33/README.md?plain=1#L121-L141

To test resumption, interrupt either side with Ctrl+C mid-session then run resume on that side to continue. A few scenarios worth covering: receiver resuming after interrupt and sender resuming after interrupt. Use history after each scenario to confirm the session state was persisted correctly.

docs for this can be seen in 7e4ffd1d3ad1165a7d9187bcacb0bdab3303dbb3

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ esplora = ["bdk_esplora", "_payjoin-dependencies"]
 rpc = ["bdk_bitcoind_rpc", "_payjoin-dependencies"] 
 
 # Internal features
-_payjoin-dependencies = ["payjoin", "reqwest", "url"]
+_payjoin-dependencies = ["payjoin", "reqwest", "url", "sqlite"]
 bip322 = ["bdk_bip322"]
 
 # Use this to consensus verify transactions at sync time
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ bdk_esplora = { version = "0.22.2", features = ["async-https", "tokio"], optiona
 bdk_kyoto = { version = "0.15.4", optional = true }
 bdk_redb = { version = "0.1.1", optional = true }
 bdk_sp = { version = "0.1.0", optional = true, git = "https://github.com/bitcoindevkit/bdk-sp", tag = "v0.1.0" }
-payjoin = { version = "=1.0.0-rc.1", features = ["v1", "v2", "io", "_test-utils"], optional = true}
-reqwest = { version = "0.13.2", default-features = false, optional = true }
+payjoin = { version = "0.25.0", features = ["v1", "v2", "io", "_test-utils"], optional = true}
+reqwest = { version = "0.13.2", default-features = false, features = ["rustls"], optional = true }
 url = { version = "2.5.8", optional = true }
 bdk_bip322 = { version = "0.1.0", optional = true }
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ And yes, it can do Taproot!!
 This crate can be used for the following purposes:
  - Instantly create a miniscript based wallet and connect to your backend of choice (Electrum, Esplora, Core RPC, Kyoto etc) and quickly play around with your own complex bitcoin scripting workflow. With one or many wallets, connected with one or many backends.
  - The `tests/integration.rs` module is used to document high level complex workflows between BDK and different Bitcoin infrastructure systems, like Core, Electrum and Lightning(soon TM).
- - Receive and send Async Payjoins. Note that even though Async Payjoin as a protocol allows the receiver and sender to go offline during the payjoin, the BDK CLI implementation currently does not support persisting.
+ - Receive and send Async Payjoins with session persistence. Sessions can be resumed if interrupted.
  - (Planned) Expose the basic command handler via `wasm` to integrate `bdk-cli` functionality natively into the web platform. See also the [playground](https://bitcoindevkit.org/bdk-cli/playground/) page.
 
 If you are considering using BDK in your own wallet project bdk-cli is a nice playground to get started with. It allows easy testnet and regtest wallet operations, to try out what's possible with descriptors, miniscript, and BDK APIs. For more information on BDK refer to the [website](https://bitcoindevkit.org/) and the [rust docs](https://docs.rs/bdk_wallet/1.0.0/bdk_wallet/index.html)
@@ -173,6 +173,33 @@ cargo run --features electrum,sp -- --network testnet4 wallet --wallet sample_wa
 It's also possible to drain all balance to a silent payment wallet by using the `--send_all` flag:
 ```shell
 cargo run --features electrum,sp -- --network testnet4 wallet --wallet sample_wallet --ext-descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" --database-type sqlite --client-type electrum --url "ssl://mempool.space:40002" create_sp_tx --send_all --to-sp <SP_CODE_1>:0
+```
+
+### Payjoin
+
+#### Payjoin Session Persistence
+
+Payjoin sessions are automatically persisted to a SQLite database (`payjoin.sqlite`) in the data directory. This allows sessions to be resumed if interrupted. Old completed sessions and stale inactive sessions older than 30 days are pruned automatically when the payjoin database is accessed.
+
+##### Resume Payjoin Sessions
+
+Resume all pending sessions:
+```
+cargo run --features rpc -- wallet --wallet <wallet_name> resume_payjoin --directory "https://payjo.in" --ohttp_relay "https://pj.bobspacebkk.com"
+```
+
+Resume a specific session by ID:
+```
+cargo run --features rpc -- wallet --wallet <wallet_name> resume_payjoin --directory "https://payjo.in" --ohttp_relay "https://pj.bobspacebkk.com" --session_id <id>
+```
+
+Sessions are processed sequentially (not concurrently) due to BDK-CLI's architecture. Each session waits up to 30 seconds for updates before timing out. If no session ID is specified, the most recent active sessions are resumed first.
+
+##### View Session History
+
+View all payjoin sessions (active and completed) and also see their status:
+```
+cargo run -- wallet --wallet <wallet_name> payjoin_history
 ```
 
 ## Justfile
--- a/src/error.rs
+++ b/src/error.rs
@@ -152,6 +152,10 @@ pub enum BDKCliError {
     #[error("Payjoin create request error: {0}")]
     PayjoinCreateRequest(#[from] payjoin::send::v2::CreateRequestError),
 
+    #[cfg(feature = "payjoin")]
+    #[error("Payjoin database error: {0}")]
+    PayjoinDb(#[from] crate::payjoin::db::Error),
+
     #[cfg(feature = "bip322")]
     #[error("BIP-322 error: {0}")]
     Bip322Error(#[from] bdk_bip322::error::Error),
@@ -181,5 +185,18 @@ impl From<bdk_redb::redb::DatabaseError> for BDKCliError {
 impl From<bdk_wallet::rusqlite::Error> for BDKCliError {
     fn from(err: bdk_wallet::rusqlite::Error) -> Self {
         BDKCliError::RusqliteError(Box::new(err))
+    }
+}
+
+#[cfg(feature = "payjoin")]
+impl<ApiErr, StorageErr, ErrorState>
+    From<payjoin::persist::PersistedError<ApiErr, StorageErr, ErrorState>> for BDKCliError
+where
+    ApiErr: std::error::Error,
+    StorageErr: std::error::Error,
+    ErrorState: std::fmt::Debug,
+{
+    fn from(e: payjoin::persist::PersistedError<ApiErr, StorageErr, ErrorState>) -> Self {
+        BDKCliError::Generic(e.to_string())
     }
 }
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -77,14 +77,7 @@ use std::convert::TryFrom;
 use std::io::Write;
 use std::path::Path;
 use std::str::FromStr;
-#[cfg(any(
-    feature = "redb",
-    feature = "compiler",
-    feature = "electrum",
-    feature = "esplora",
-    feature = "cbf",
-    feature = "rpc"
-))]
+#[cfg(any(feature = "redb", feature = "compiler"))]
 use std::sync::Arc;
 
 #[cfg(feature = "bip322")]
@@ -100,9 +93,8 @@ use bdk_bip322::{BIP322, MessageProof, MessageVerificationResult};
 ))]
 use {
     crate::commands::OnlineWalletSubCommand::*,
-    crate::payjoin::{PayjoinManager, ohttp::RelayManager},
+    crate::payjoin::PayjoinManager,
     bdk_wallet::bitcoin::{Transaction, consensus::Decodable, hex::FromHex},
-    std::sync::Mutex,
 };
 #[cfg(feature = "esplora")]
 use {crate::utils::BlockchainClient::Esplora, bdk_esplora::EsploraAsyncExt};
@@ -863,6 +855,8 @@ pub(crate) async fn handle_online_wallet_subcommand(
     wallet: &mut Wallet,
     client: &BlockchainClient,
     online_subcommand: OnlineWalletSubCommand,
+    datadir: Option<std::path::PathBuf>,
+    wallet_name: &str,
 ) -> Result<String, Error> {
     match online_subcommand {
         FullScan {
@@ -989,8 +983,7 @@ pub(crate) async fn handle_online_wallet_subcommand(
             ohttp_relay,
             max_fee_rate,
         } => {
-            let relay_manager = Arc::new(Mutex::new(RelayManager::new()));
-            let mut payjoin_manager = PayjoinManager::new(wallet, relay_manager);
+            let mut payjoin_manager = PayjoinManager::new(wallet, datadir.clone(), wallet_name)?;
             return payjoin_manager
                 .receive_payjoin(amount, directory, max_fee_rate, ohttp_relay, client)
                 .await;
@@ -1000,12 +993,22 @@ pub(crate) async fn handle_online_wallet_subcommand(
             ohttp_relay,
             fee_rate,
         } => {
-            let relay_manager = Arc::new(Mutex::new(RelayManager::new()));
-            let mut payjoin_manager = PayjoinManager::new(wallet, relay_manager);
+            let mut payjoin_manager = PayjoinManager::new(wallet, datadir.clone(), wallet_name)?;
             return payjoin_manager
                 .send_payjoin(uri, fee_rate, ohttp_relay, client)
                 .await;
         }
+        ResumePayjoin {
+            directory,
+            ohttp_relay,
+            session_id,
+        } => {
+            let mut payjoin_manager = PayjoinManager::new(wallet, datadir, wallet_name)?;
+            return payjoin_manager
+                .resume_payjoins(directory, ohttp_relay, session_id, client)
+                .await;
+        }
+        PayjoinHistory => crate::payjoin::PayjoinManager::history(datadir, wallet_name),
     }
 }
 
@@ -1511,14 +1514,14 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
             feature = "rpc"
         ))]
         CliSubCommand::Wallet {
-            wallet,
+            wallet: wallet_name,
             subcommand: WalletSubCommand::OnlineWalletSubCommand(online_subcommand),
         } => {
-            let home_dir = prepare_home_dir(cli_opts.datadir)?;
+            let home_dir = prepare_home_dir(cli_opts.datadir.clone())?;
 
-            let (wallet_opts, network) = load_wallet_config(&home_dir, &wallet)?;
+            let (wallet_opts, network) = load_wallet_config(&home_dir, &wallet_name)?;
 
-            let database_path = prepare_wallet_db_dir(&home_dir, &wallet)?;
+            let database_path = prepare_wallet_db_dir(&home_dir, &wallet_name)?;
 
             #[cfg(any(feature = "sqlite", feature = "redb"))]
             let result = {
@@ -1532,13 +1535,16 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     }
                     #[cfg(feature = "redb")]
                     DatabaseType::Redb => {
-                        let wallet_name = &wallet_opts.wallet;
+                        let redb_store_wallet_name = &wallet_opts.wallet;
                         let db = Arc::new(bdk_redb::redb::Database::create(
                             home_dir.join("wallet.redb"),
                         )?);
                         let store = RedbStore::new(
                             db,
-                            wallet_name.as_deref().unwrap_or("wallet").to_string(),
+                            redb_store_wallet_name
+                                .as_deref()
+                                .unwrap_or("wallet")
+                                .to_string(),
                         )?;
                         log::debug!("Redb database opened successfully");
                         Persister::RedbStore(store)
@@ -1553,6 +1559,8 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     &mut wallet,
                     &blockchain_client,
                     online_subcommand,
+                    cli_opts.datadir.clone(),
+                    &wallet_name,
                 )
                 .await?;
                 wallet.persist(&mut persister)?;
@@ -1563,8 +1571,14 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                 let mut wallet = new_wallet(network, wallet_opts)?;
                 let blockchain_client =
                     crate::utils::new_blockchain_client(wallet_opts, &wallet, database_path)?;
-                handle_online_wallet_subcommand(&mut wallet, &blockchain_client, online_subcommand)
-                    .await?
+                handle_online_wallet_subcommand(
+                    &mut wallet,
+                    &blockchain_client,
+                    online_subcommand,
+                    cli_opts.datadir.clone(),
+                    &wallet_name,
+                )
+                .await?
             };
             Ok(result)
         }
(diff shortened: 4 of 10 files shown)


src/handlers.rs
```diff
@@ -594,6 +594,16 @@ pub fn handle_offline_wallet_subcommand(
     }
 }
 
+#[cfg(feature = "payjoin")]
+pub fn open_payjoin_db(
+    datadir: Option<std::path::PathBuf>,
+) -> Result<std::sync::Arc<crate::payjoin::db::Database>, Error> {
+    use crate::payjoin::db::{DB_FILENAME, Database};
+    let home_dir = prepare_home_dir(datadir)?;
+    let db_path = home_dir.join(DB_FILENAME);
```
tvpeter: I think it will be more reasonable if the db path includes the specific wallet that is used instead of having the payjoin db at the directory level

src/handlers.rs
```diff
@@ -594,6 +594,16 @@ pub fn handle_offline_wallet_subcommand(
     }
 }
 
+#[cfg(feature = "payjoin")]
```
tvpeter: Since `sqlite` is also a feature, this should be available only when both `sqlite` and `payjoin` features are enabled

src/handlers.rs
```diff
@@ -594,6 +594,16 @@ pub fn handle_offline_wallet_subcommand(
...
+pub fn open_payjoin_db(
```
tvpeter: The fn will be more suitable in the utils file

src/payjoin/mod.rs
```diff
@@ -807,4 +925,249 @@ impl<'a> PayjoinManager<'a> {
...
+        let db = self.db.clone();
+        let mut recv_session_ids = db.get_recv_session_ids()?;
+        let mut send_session_ids = db.get_send_session_ids()?;
+
+        if let Some(session_id) = session_id {
+            recv_session_ids.retain(|id| id.as_i64() == session_id);
+            send_session_ids.retain(|id| id.as_i64() == session_id);
+
+            if recv_session_ids.is_empty() && send_session_ids.is_empty() {
+                return Ok(serde_json::to_string_pretty(&json!({
```
tvpeter, line 862: Instead of erroring out when no session id is provided, I think it will be a better flow to select the most recent session id and resume that session.

tvpeter (changes requested):
Thank you @Mshehu5 for working on this feature.

Below are some of the observations that I think will make the implementation better:

- I noticed that the implementation used only the `sqlite` db, and since `sqlite` is a db option in the project, I think it will be ideal if you also include implementation for `redb` db so users are free to choose any to use.
- The implementation does not provide for pruning/cleanup of data. Given that the db will tend to grow linearly and become sizable over time, I think it will be great to consider adding a pruning subcommand or mechanism that deletes irrelevant data.
- Since the `save_event` is generated at multiple transitions in a session, I think it will be great if you consider grouping such updates in a db transaction to ensure that entries are saved successfully.

I have also left some comments in the code.

Thank you.

src/payjoin/mod.rs
Mshehu5: This is particular code is for resuming a single session by session ID for the resume payjoin commad we already do this, it resumes the most reccent session, resuming the latest session instead here might make the user think session exists and is being resumed

Mshehu5:
Thank you for the review!

Just wanted some clarifications 
> The implementation does not provide for pruning/cleanup of data. Given that the db will tend to grow linearly and become sizable over time, I think it will be great to consider adding a pruning subcommand or mechanism that deletes irrelevant data.

For this feature, how would you like it to be implemented? For example should we allow selecting a specific date or session to start pruning from? Or should it apply to all sessions older than a certain date? Alternatively should it target only expired or unsuccessful sessions?

> I noticed that the implementation used only the sqlite db, and since sqlite is a db option in the project, I think it will be ideal if you also include implementation for redb db so users are free to choose any to use.

Also for the redb implementation most of the existing integrations have been built around SQLite so it made sense as the initial approach for this implementation. The Redb support may require some additional consideration and it would be good to align with the PayJoin team to ensure the approach is correct and consistent.
Given that would you prefer the Redb implementation to be included in this PR or would it be acceptable to handle it in a follow-up?

It would also be helpful to understand what you consider a hard requirement for this PR versus what can be iterated on in subsequent changes.

va-an (commented):
Hi @Mshehu5, thanks for this!

Could you update payjoin dependency to `1.0.0-rc.2` and run `just pre-push`?

I've got some errors from `just pre-push` like:
```sh
error[E0061]: this method takes 1 argument but 2 arguments were supplied
    --> src/payjoin/mod.rs:691:30
     |
 691 |   ...                   .check_payment(
     |                          ^^^^^^^^^^^^^
...
 704 | / ...                       |outpoint| {
 705 | | ...                           let utxo = self.wallet.get_utxo(outpoint);
 706 | | ...                           match utxo {
 707 | | ...                               Some(_) => Ok(false),
...    |
 710 | | ...                       }
     | |___________________________- unexpected argument #2 of type `{closure@src/payjoin/mod.rs:704:33: 704:43}`
```

Mshehu5:
@va-an Thanks for the review! I bumped Payjoin to 1.0.0-rc.2 and just pre-push passes locally. Mind giving it a try on your end? Would appreciate your feedback

Mshehu5:
@tvpeter Would really appreciate your feedback on this: https://github.com/bitcoindevkit/bdk-cli/pull/242#issuecomment-4121462117

It would help move this PR forward.

tvpeter:
> For this feature, how would you like it to be implemented? For example should we allow selecting a specific date or session to start pruning from? Or should it apply to all sessions older than a certain date? Alternatively should it target only expired or unsuccessful sessions?

I think it should apply to all sessions within a specific timeframe, say anything older than 30 days, and if it can be implemented without adding another command, that would be great. This can even go into another PR.

> Also for the redb implementation most of the existing integrations have been built around SQLite so it made sense as the initial approach for this implementation. The Redb support may require some additional consideration and it would be good to align with the PayJoin team to ensure the approach is correct and consistent. Given that would you prefer the Redb implementation to be included in this PR or would it be acceptable to handle it in a follow-up?

That is still fine. It can form another PR, but it's one of the things you can consider adding.

> It would also be helpful to understand what you consider a hard requirement for this PR versus what can be iterated on in subsequent changes.

The current implementation covering SQLite is good enough (and the most important). Those others can serve as improvements to the persistence feature.

src/payjoin/mod.rs
tvpeter: Alright, that's fine

src/payjoin/db.rs
```diff
@@ -109,6 +110,78 @@ impl Database {
...
+                "DELETE FROM receive_session_events WHERE session_id = ?1",
+                params![session_id],
+            )?;
+            tx.execute(
+                "DELETE FROM receive_sessions WHERE session_id = ?1",
+                params![session_id],
+            )?;
+        }
+
+        tx.commit()?;
```
tvpeter, line 238: you can add `db vacuum` command after the commit to reclaim actual disk space.

README.md
```diff
@@ -150,6 +150,31 @@ cargo run --features rpc -- wallet --wallet payjoin_wallet2 balance
 cargo run --features rpc -- wallet --wallet payjoin_wallet2 send_payjoin --ohttp_relay "https://pj.bobspacebkk.com" --ohttp_relay "https://pj.benalleng.com" --fee_rate 1 --uri "<URI>"
 ```
 
+### Payjoin Session Persistence
```
tvpeter: While this is the right thing to do, pls can you make **Payjoin** as a subheading starting from [here](https://github.com/bitcoindevkit/bdk-cli/pull/242/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L131) and let it have all the below as its subheadings.

tvpeter (commented):
Hi @Mshehu5 ,

Thank you so much for continuing to work on this feature. I left some minor comments. After you address them, I will do a final round of testing. Overall, I feel the PR is ready to go.

Well done.

tvpeter (approved):
ACK d22d8b05c01a88e29852444ab577c6f02f315f17

I was unable to test the entire payjoin flow (due to a persistent network error), but I could see the `resume` and `history` commands working as expected.

Weldone @Mshehu5

tvpeter:
@Mshehu5 please rebase

Mshehu5:
Hello @notmandatory A review/feedback from you on this PR will really help push this forward

DanGould:
@evanlinjin this might be a PR for us to look at together this week

src/payjoin/mod.rs
```diff
@@ -176,28 +245,64 @@ impl<'a> PayjoinManager<'a> {
                     ));
                 }
 
-                // TODO: Implement proper persister.
-                let persister =
-                    payjoin::persist::NoopSessionPersister::<SenderSessionEvent>::default();
+                use payjoin::send::v2::replay_event_log as replay_sender_event_log;
```
vadim-anfv: This import is already on line 22.

```suggestion
```

src/payjoin/db.rs
```diff
@@ -0,0 +1,475 @@
...
+/// Wrapper type for session IDs
+#[derive(Debug, Clone)]
+pub struct SessionId(i64);
+
+impl core::ops::Deref for SessionId {
+    type Target = i64;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
vadim-anfv: `impl Deref` is only used to unwrap the inner `i64` for `params!`. Per [C-DEREF](https://rust-lang.github.io/api-guidelines/predictability.html#c-deref), `Deref` should be implemented _only_ for smart pointers - not for newtypes.

Cleaner: `impl ToSql for SessionId` (delegating to `self.0`), then `params![self.session_id, ...]` works directly. Drop the `Deref` impl and remove `*` from every `params![*session_id, ...]` call site.

```suggestion
impl ToSql for SessionId {
    fn to_sql(&self) -> bdk_wallet::rusqlite::Result<ToSqlOutput<'_>> {
        self.0.to_sql()
    }
}
```

src/payjoin/mod.rs
vadim-anfv: Stale TODO.

```suggestion
```

src/handlers.rs
```diff
@@ -782,11 +785,25 @@ pub(crate) async fn handle_online_wallet_subcommand(
...
+            directory,
+            ohttp_relay,
+            session_id,
+        } => {
+            let relay_manager = Arc::new(Mutex::new(RelayManager::new()));
+            let db = open_payjoin_db(datadir, &wallet_name)?;
+            let mut payjoin_manager = PayjoinManager::new(wallet, relay_manager, db);
+            return payjoin_manager
+                .resume_payjoins(directory, ohttp_relay, session_id, client)
+                .await;
```
vadim-anfv: The same 3-line boilerplate is repeated before every `PayjoinManager::new`:

```rust
let relay_manager = Arc::new(Mutex::new(RelayManager::new()));
let db = open_payjoin_db(datadir.clone(), &wallet_name)?;
let mut payjoin_manager = PayjoinManager::new(wallet, relay_manager, db);
```

Both `db` and `relay_manager` are used only by `PayjoinManager` - callers don't touch them. Move construction inside `new`:

```rust
pub fn new(
    wallet: &'a mut Wallet,
    datadir: Option<PathBuf>,
    wallet_name: &str,
) -> Result<Self, Error> {
    let db = open_payjoin_db(datadir, wallet_name)?;
    let relay_manager = Arc::new(Mutex::new(RelayManager::new()));
    Ok(Self { wallet, relay_manager, db })
}
```

Callers collapse to one line.

src/utils.rs
```diff
@@ -152,6 +152,18 @@ pub(crate) fn prepare_wallet_db_dir(
 
     Ok(dir)
 }
+#[cfg(feature = "payjoin")]
+pub fn open_payjoin_db(
```
vadim-anfv: This function should be moved to `payjoin/db.rs`.

va-an (changes requested):
Tried end-to-end on regtest with `pj.bobspacebkk.com`. The URI is generated and the persister state is saved correctly, but the first poll request fails:

```
-> % RUST_LOG=debug cargo run --features rpc -- \
    wallet --wallet payjoin_wallet1 \
    receive_payjoin \
    --amount 400000 \
    --max_fee_rate 1000 \
    --directory "https://payjo.in" \
    --ohttp_relay "https://pj.bobspacebkk.com"
...
[2026-05-26T20:05:14Z DEBUG bdk_cli] network: Testnet
[2026-05-26T20:05:14Z DEBUG bdk_cli::handlers] Sqlite database opened successfully
[2026-05-26T20:05:14Z DEBUG reqwest::connect] starting new connection: https://payjo.in/
[2026-05-26T20:05:14Z DEBUG reqwest::connect] proxy(https://pj.bobspacebkk.com/) intercepts 'https://payjo.in/'
Request Payjoin by sharing this Payjoin Uri:
bitcoin:bcrt1qn0wg553k3yfeg56535tej03wxv5yftzx4gpvpe?amount=0.004&pjos=0&pj=HTTPS://PAYJO.IN/SM6XG5LDPYGLC%23EX10D8PW6S-OH1QYPFLM8XL59R0XV4VGPLS7FRDSSM4TUXL07TXCWC4S0GLVLNK2SE4NQ-RK1QFF9Z8Y93K4TUVGJAHR9AX27SDPSJ88FY8HAC7SQ9DFCMCTJ5LSTZ
[2026-05-26T20:05:15Z DEBUG reqwest::connect] starting new connection: https://pj.bobspacebkk.com/
[2026-05-26T20:05:15Z ERROR bdk_cli] Reqwest error: error sending request for url (https://pj.bobspacebkk.com/https://payjo.in/)

```

After downgrading `Cargo.toml` to `reqwest = "0.12.23"` (the version used by the reference `payjoin-cli`), the polling succeeds and the CLI hangs waiting for the sender as expected:

```
-> % RUST_LOG=debug cargo run --features rpc -- \
    wallet --wallet payjoin_wallet1 \
    receive_payjoin \
    --amount 400000 \
    --max_fee_rate 1000 \
    --directory "https://payjo.in" \
    --ohttp_relay "https://pj.bobspacebkk.com"
...
[2026-05-26T20:04:24Z DEBUG bdk_cli] network: Testnet
[2026-05-26T20:04:24Z DEBUG bdk_cli::handlers] Sqlite database opened successfully
[2026-05-26T20:04:24Z DEBUG reqwest::connect] starting new connection: https://payjo.in/
[2026-05-26T20:04:24Z DEBUG reqwest::connect] proxy(https://pj.bobspacebkk.com/) intercepts 'https://payjo.in/'
Request Payjoin by sharing this Payjoin Uri:
bitcoin:bcrt1qn0wg553k3yfeg56535tej03wxv5yftzx4gpvpe?amount=0.004&pjos=0&pj=HTTPS://PAYJO.IN/7X88PY7FGVM07%23EX1F98PW6S-OH1QYPFLM8XL59R0XV4VGPLS7FRDSSM4TUXL07TXCWC4S0GLVLNK2SE4NQ-RK1Q2DC9EW9ECFDEZJ964H9ELY83F8HM7P3AFSJSFNW6VY9D3RPXE80X
[2026-05-26T20:04:25Z DEBUG reqwest::connect] starting new connection: https://pj.bobspacebkk.com/
^C
```

The reqwest 0.13.2 bump wasn't introduced by this PR (likely came in via a rebase on `master`), but as it stands the code doesn't work end-to-end.

src/error.rs
```diff
@@ -176,3 +176,64 @@ impl From<bdk_wallet::rusqlite::Error> for BDKCliError {
         BDKCliError::RusqliteError(Box::new(err))
     }
 }
+
+/// Error type for payjoin database operations
+#[cfg(feature = "payjoin")]
+#[derive(Debug)]
+pub enum PayjoinDbError {
```
DanGould: Unless these new error types are needed at this top level visibility, I'd move them into the `payjoin::db` module. it's in the name. `payjoin::db::Error` seems like the appropriate place/name for this type.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ bdk_esplora = { version = "0.22.1", features = ["async-https", "tokio"], optiona
 bdk_kyoto = { version = "0.15.4", optional = true }
 bdk_redb = { version = "0.1.1", optional = true }
 shlex = {  version = "1.3.0", optional = true }
-payjoin = { version = "=1.0.0-rc.1", features = ["v1", "v2", "io", "_test-utils"], optional = true}
+payjoin = { version = "1.0.0-rc.2", features = ["v1", "v2", "io", "_test-utils"], optional = true}
```
DanGould: Using the latest, payjoin version="0.25.0" for this and keep updating proper 1.0 is going to stay closer to the released api.

src/payjoin/mod.rs
```diff
@@ -71,8 +88,8 @@ impl<'a> PayjoinManager<'a> {
 
         let ohttp_keys =
             fetch_ohttp_keys(ohttp_relays, &directory, self.relay_manager.clone()).await?;
-        // TODO: Implement proper persister.
-        let persister = payjoin::persist::NoopSessionPersister::<ReceiverSessionEvent>::default();
+
+        let persister = crate::payjoin::db::ReceiverPersister::new(self.db.clone())?;
```
DanGould, line 140: Seems like the `PayjoinManager` could encapsulate the `ReceiverPersister` around the db itself. Why not?

src/payjoin/mod.rs
```diff
@@ -175,10 +192,45 @@ impl<'a> PayjoinManager<'a> {
...
+                            .expect("Receiver pubkey should exist if session id exists");
+                        if session_receiver_pubkey == *receiver_pubkey {
+                            let sender_persister =
+                                SenderPersister::from_id(self.db.clone(), session_id);
+                            let (send_session, _) = replay_sender_event_log(&sender_persister)
+                                .map_err(|e| {
+                                    Error::Generic(format!(
+                                        "Failed to replay sender event log: {:?}",
+                                        e
+                                    ))
```
DanGould: Why generic error instead of panic if session nonexistence otherwise panics?

src/payjoin/mod.rs
```diff
@@ -545,22 +580,21 @@ impl<'a> PayjoinManager<'a> {
...
     ) -> Result<(), Error> {
-        let (req, ctx) = receiver.create_post_request(
-            self.relay_manager
-                .lock()
-                .expect("Lock should not be poisoned")
-                .get_selected_relay()
-                .expect("A relay should already be selected")
+        let (req, ctx) = receiver
+            .create_post_request(
+                self.unwrap_relay_or_else_fetch(vec![], None::<&str>)
```
DanGould: No relay no directory? What's the target of this request? Why not use the fetch function directly if not relay manager?

src/payjoin/mod.rs
```diff
@@ -729,31 +773,53 @@ impl<'a> PayjoinManager<'a> {
         }
     }
 
+    async fn unwrap_relay_or_else_fetch(
+        &self,
+        ohttp_relays: Vec<url::Url>,
```
DanGould: why even pass in ohttp_relays if there's a relay manager? Seems like there are 2 ideas here and one needs to be picked and the other removed.

src/payjoin/mod.rs
```diff
@@ -729,31 +773,53 @@ impl<'a> PayjoinManager<'a> {
...
+        }
+    }
+
     async fn post_original_proposal(
         &self,
         sender: Sender<WithReplyKey>,
-        relay: impl payjoin::IntoUrl,
         persister: &impl SessionPersister<SessionEvent = SenderSessionEvent>,
         blockchain_client: &BlockchainClient,
+        relay: String,
```
DanGould: This change that loosens the contract doesn't make sense to me. Still want to enforce that the param is of the correct type even if you convert it to a string in the method body.

src/payjoin/mod.rs
```diff
@@ -807,4 +873,252 @@ impl<'a> PayjoinManager<'a> {
...
+        println!("Resuming {} payjoin session(s)...\n", total_sessions);
+
+        // Resume receiver sessions
+        for session_id in recv_session_ids {
+            let persister = ReceiverPersister::from_id(db.clone(), session_id.clone());
+            match replay_receiver_event_log(&persister) {
+                Ok((receiver_state, _)) => {
+                    println!("Resuming receiver session {}", session_id);
+                    match tokio::time::timeout(
+                        std::time::Duration::from_secs(30),
```
DanGould, line 896: why time out after 30 seconds? Seems brief since payjoins are long-running and don't expire by default for 24h

src/handlers.rs
```diff
@@ -60,14 +60,7 @@ use std::convert::TryFrom;
...
 use std::str::FromStr;
-#[cfg(any(
-    feature = "redb",
-    feature = "compiler",
-    feature = "electrum",
-    feature = "esplora",
-    feature = "cbf",
-    feature = "rpc"
-))]
+#[cfg(any(feature = "redb", feature = "compiler"))]
```
DanGould, line 80: This seems like it's touching things outside of the scope of this PR. Are you sure these feature gates can be removed?

src/handlers.rs
```diff
@@ -1296,6 +1300,8 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     &mut wallet,
                     &blockchain_client,
                     online_subcommand,
+                    cli_opts.datadir.clone(),
+                    wallet_name.clone(),
```
DanGould: passing `wallet` and `wallet_name` which are cloned versions of the same type to this function seems very definitely like a hack. What the heck is going on here?

src/payjoin/mod.rs
```diff
@@ -232,14 +286,14 @@ impl<'a> PayjoinManager<'a> {
...
-                        ))
-                    })?;
+                    let sender = payjoin::send::v2::SenderBuilder::new(original_psbt.clone(), uri)
+                        .build_recommended(fee_rate)?
+                        .save(&persister)
+                        .map_err(|e| {
+                            Error::Generic(format!(
+                                "Failed to save the Payjoin v2 sender in the persister: {e}"
+                            ))
+                        })?;
```
DanGould, line 274: This kind of spacing-only change makes me think you haven't reviewed this line-by-line yourself before asking for review here. Please make sure these kinds of formatting issues you can find yourself are addressed beforehand.

DanGould (commented):
Went through this in prep for mob review, although I don't have.

The lack of any explained decision motivation in the commit messages is the fundamental to change first whenever asking for review. Whoever reads the history can see what is done. The commit message exists to tell the history of WHY you made changes. Tell me what you considered so I don't have to imagine.

Some things we'll talk about:

- appropriate use of modules
- rebasing
- Encapsulating payjoin business from cli business
- using the right version
- The right abstraction level for db vs a payjoin-specific storage/manager type

src/payjoin/db.rs
```diff
@@ -0,0 +1,486 @@
...
+fn now() -> i64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_secs() as i64
+}
+
+pub struct Database {
+    conn: Mutex<Connection>,
+}
```
evanlinjin, line 87: This `struct` does not earn it's keep and neither does the `Mutex`.

I think it would make more sense just to have separate functions instead of methods on `Database` (that doesn't need to be there).

The `...Persister` types should just be:

```rust
stuct XxxPersister {
    conn: Rc<Connection>,
    session_id: SessionId,
} 
```

evanlinjin (changes requested):

src/payjoin/mod.rs
DanGould: This is how the reference does it https://github.com/payjoin/rust-payjoin/issues/1582, though I note the payjoin-cli reference has no `vec![]` which adds further confusion to my eye

Mshehu5:
@va-an Thank you for the review. The reqwest issue is fixed in 7ee0a71b113c46b4bda8cf927af37859dfdeb550. The regression was caused by a separate merged PR that updated reqwest without enabling rustls. I have verified the fix locally and it should be ready for you to retest on your end.

src/payjoin/mod.rs
Mshehu5: vec![] here is not the request target. In this receiver flow the relay was already selected earlier when we fetched OHTTP keys and that relay is stored in relay_manager so this call just reuses it.

The reason payjoin-cli does not need the extra Vec is that it reads relay candidates from config inside the helper. In bdk-cli, relay candidates come from CLI input so the shared helper takes them as a parameter even though this specific receiver call site does not use them.

I could break down or add a function to just do select relay instead of trying to use unwrap_relay_or_else_fetch and having to settle the vec![] for thr function signature

src/payjoin/mod.rs
Mshehu5: saw it in the reference implementation so I just went that direction. looked like the preferred 
https://github.com/payjoin/rust-payjoin/blob/fb884d11996acea9d9bd5573b32b549b1bfdc334/payjoin-cli/src/app/v2/mod.rs#L230

src/handlers.rs
Mshehu5: This is related to https://github.com/bitcoindevkit/bdk-cli/pull/242#discussion_r3305804299 . handlers.rs no longer constructs Arc<Mutex<RelayManager>> or opens the Payjoin DB before each PayjoinManager::new call. That setup now lives inside PayjoinManager::new.

After that refactor, Arc is no longer used in handlers.rs for rpc/cbf/electrum/esplora and Mutex is not used there at all. The remaining Arc uses are only under redb and compiler so the narrowed cfg matches the actual users.

src/payjoin/mod.rs
Mshehu5: I forgot to talk about this during mob review because it was one of the tradeoff/design decisions in this PR.

This timeout is for a single session resume attempt not the session lifetime. I kept resume synchronous unlike payjoin-cli because the wallet in this codebase is currently handled as a single mutable resource. Making resume concurrent would require a broader refactor around shared wallet access and locking which felt too large (may change alot of things and how the maintainers intend codebase to be ) and beyond the scope of this PR (This is already a big PR ).
Given that, the 30-second outer timeout is there to prevent one resumed session from hanging indefinitely and blocking the rest of the queue. The payjoin session itself is still persisted and can be resumed again later this does not replace the protocol’s longer expiration window.

notmandatory (approved):
tACK bdb7f3869a4ce17005b935dd3470c623e9e4fbd8

Thanks for the detailed test steps, very helpful for manual testing. I verified canceling receive and separately sender flow works as expected.

Any other feature enhancements such as using a DB transaction for atomic updates, or improved error types I recommend making new issues for. These sort of small improvements would be better made after #278 and in a future milestone. I also don't think we need this to support the redb persistence right now. If someone needs it they can implement it in a new PR.

tvpeter:
@Mshehu5 pls rebase

PR #248, opened 2026-03-13
Update Wallet to 2.4.0 and other deps

### LINKED ISSUE #243: Update out of date dependencies
With a local installation, we see warnings about deprecated structs:
```bash
-> % cargo install --path . --all-features
...
warning: use of deprecated enum `bdk_wallet::signer::SignerError`: PSBT signing was moved to `bitcoin::psbt` module
  --> src/error.rs:79:45
   |
79 |     SignerError(#[from] bdk_wallet::signer::SignerError),
   |                                             ^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
...
```

This happens because local installation does not take `Cargo.lock` into account and pulls the latest compatible dependency versions. In particular, the warnings come from bdk_wallet 2.3.0.

**What have to be done:**
- update bdk_wallet to version 2.3.0
- fix the code where required to address deprecations
- update other outdated dependencies as well (see: https://deps.rs/repo/github/bitcoindevkit/bdk-cli).

tvpeter:
### Description
This PR updates the Wallet API to v2.4.0 and other dependencies. It also adds WalletEvent to both sync and full_scan for all backends.

Fixes #243 

## Changelog notice
- Update bdk_wallet to v2.4.0
- Add WalletEvents to all backends Esplora, Electrum, Rpc and CBF.
- Add `print_wallet_events` function to print wallet events to the terminal during sync
- Update bdk_bitcoind_rpc to v0.22.0
- Update bdk_electrum to v0.24.0
- Update bdk_kyoto to v0.15.4
- Update bdk_esplora to v0.22.2

#### All Submissions:

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -633,7 +633,7 @@ pub(crate) async fn handle_online_wallet_subcommand(
                     client
                         .populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-                    let update = client.full_scan(request, _stop_gap, batch_size, false)?;
+                    let update = client.full_scan(request, _stop_gap, *batch_size, false)?;
                     let wallet_events = wallet.apply_update_events(update)?;
                     print_wallet_events(&wallet_events);
                 }
@@ -1537,7 +1537,7 @@ pub async fn sync_wallet(client: &BlockchainClient, wallet: &mut Wallet) -> Resu
             // already have.
             client.populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-            let update = client.sync(request, batch_size, false)?;
+            let update = client.sync(request, *batch_size, false)?;
             let wallet_events = wallet
                 .apply_update_events(update)
                 .map_err(|e| Error::Generic(e.to_string()))?;
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -633,8 +633,9 @@ pub(crate) async fn handle_online_wallet_subcommand(
                     client
                         .populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-                    let update = client.full_scan(request, _stop_gap, *batch_size, false)?;
-                    wallet.apply_update(update)?;
+                    let update = client.full_scan(request, _stop_gap, batch_size, false)?;
+                    let wallet_events = wallet.apply_update_events(update)?;
+                    print_wallet_events(&wallet_events);
                 }
                 #[cfg(feature = "esplora")]
                 Esplora {
@@ -645,7 +646,8 @@ pub(crate) async fn handle_online_wallet_subcommand(
                         .full_scan(request, _stop_gap, *parallel_requests)
                         .await
                         .map_err(|e| *e)?;
-                    wallet.apply_update(update)?;
+                    let wallet_events = wallet.apply_update_events(update)?;
+                    print_wallet_events(&wallet_events);
                 }
 
                 #[cfg(feature = "rpc")]
@@ -1535,10 +1537,12 @@ pub async fn sync_wallet(client: &BlockchainClient, wallet: &mut Wallet) -> Resu
             // already have.
             client.populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-            let update = client.sync(request, *batch_size, false)?;
-            wallet
-                .apply_update(update)
-                .map_err(|e| Error::Generic(e.to_string()))
+            let update = client.sync(request, batch_size, false)?;
+            let wallet_events = wallet
+                .apply_update_events(update)
+                .map_err(|e| Error::Generic(e.to_string()))?;
+            print_wallet_events(&wallet_events);
+            Ok(())
         }
         #[cfg(feature = "esplora")]
         Esplora {
@@ -1549,9 +1553,11 @@ pub async fn sync_wallet(client: &BlockchainClient, wallet: &mut Wallet) -> Resu
                 .sync(request, *parallel_requests)
                 .await
                 .map_err(|e| *e)?;
-            wallet
-                .apply_update(update)
-                .map_err(|e| Error::Generic(e.to_string()))
+            let wallet_events = wallet
+                .apply_update_events(update)
+                .map_err(|e| Error::Generic(e.to_string()))?;
+            print_wallet_events(&wallet_events);
+            Ok(())
         }
         #[cfg(feature = "rpc")]
         RpcClient { client } => {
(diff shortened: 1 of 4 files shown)


src/handlers.rs
```diff
@@ -632,8 +632,9 @@ pub(crate) async fn handle_online_wallet_subcommand(
                     client
                         .populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-                    let update = client.full_scan(request, _stop_gap, *batch_size, false)?;
-                    wallet.apply_update(update)?;
+                    let update = client.full_scan(request, _stop_gap, batch_size, false)?;
```
vadim-anfv:
```suggestion
                    let update = client.full_scan(request, _stop_gap, *batch_size, false)?;

```

src/handlers.rs
```diff
@@ -1524,10 +1526,12 @@ pub async fn sync_wallet(client: &BlockchainClient, wallet: &mut Wallet) -> Resu
             // already have.
             client.populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
 
-            let update = client.sync(request, *batch_size, false)?;
-            wallet
-                .apply_update(update)
-                .map_err(|e| Error::Generic(e.to_string()))
+            let update = client.sync(request, batch_size, false)?;
```
vadim-anfv:
```suggestion
            let update = client.sync(request, *batch_size, false)?;
```

va-an (changes requested):
Blocked by #252.

va-an:
@tvpeter #252 is merged now, so this should be fine now. 
Could you rebase on master? Other PRs will likely need to rebase on top of this once it's merged.

va-an:
Thanks for the rebase, @tvpeter!

Unfortunately, CI is now failing because `bdk_wallet` 2.3.0 introduced a deprecation that causes warnings-as-errors to fail. The [revert](https://github.com/bitcoindevkit/bdk_wallet/pull/405) has already been merged but [2.4.0](https://github.com/bitcoindevkit/bdk_wallet/issues/426) hasn't been released yet.

Pinning back to 2.1.0 won't work either - this branch already uses `apply_update_events` which was introduced in 2.3.0, so downgrading breaks the build:

```
error[E0599]: no method named `apply_update_events` found for mutable reference `&mut Wallet`
```

So it looks like we're stuck waiting for 2.4.0 to be released. Until then, CI will fail on this PR (and on any PR that rebases after merging this one). 
We could just keep this in mind during reviews and merges - the CI failure is a known issue, not a code problem.

src/handlers.rs
```diff
@@ -1536,9 +1538,11 @@ pub async fn sync_wallet(client: &BlockchainClient, wallet: &mut Wallet) -> Resu
...
 
             let update = client.sync(request, *batch_size, false)?;
-            wallet
-                .apply_update(update)
-                .map_err(|e| Error::Generic(e.to_string()))
+            let wallet_events = wallet
+                .apply_update_events(update)
+                .map_err(|e| Error::Generic(e.to_string()))?;
+            print_wallet_events(&wallet_events);
+            Ok(())
```
vadim-anfv, line 1852: We could use [tap](https://crates.io/crates/tap) for convenience:

```suggestion
            wallet
                .apply_update_events(update)
                .map_err(|e| Error::Generic(e.to_string()))?
                .tap(print_wallet_events);

            Ok(())
```

src/utils.rs
```diff
@@ -663,3 +665,45 @@ pub fn load_wallet_config(
 
     Ok((wallet_opts, network))
 }
+#[cfg(any(feature = "electrum", feature = "esplora",))]
+pub fn print_wallet_events(events: &Vec<WalletEvent>) {
```
vadim-anfv:
```suggestion
pub fn print_wallet_events(events: &[WalletEvent]) {
```

va-an (approved):
utACK 4a41057fbe2a9462b2aea3c3849de6ebed2b00aa

tvpeter:
> Unfortunately, CI is now failing because `bdk_wallet` 2.3.0 introduced a deprecation that causes warnings-as-errors to fail. 

Yes, I updated the warnings to be treated as errors in this commit 8d20cf270d7e9fee71857135cb94b0d22a4b6565 because many Clippy warnings were being neglected and had to be merged into master. It's not coming from bdk_wallet

va-an:
> Yes, I updated the warnings to be treated as errors in this commit [8d20cf2](https://github.com/bitcoindevkit/bdk-cli/commit/8d20cf270d7e9fee71857135cb94b0d22a4b6565) because many Clippy warnings were being neglected and had to be merged into master. It's not coming from bdk_wallet

The `deny(warnings)` change is great, glad you added it - lax warnings tend to pile up fast. The issue here is a different thing though: `bdk_wallet` 2.3.0 itself emits a deprecation warning on its own API (`apply_update`), and that warning hits the `deny(warnings)` flag. It'll be resolved once 2.4.0 is released (bitcoindevkit/bdk_wallet#405).

notmandatory:
> Unfortunately, CI is now failing because `bdk_wallet` 2.3.0 introduced a deprecation that causes warnings-as-errors to fail. The [revert](https://github.com/bitcoindevkit/bdk_wallet/pull/405) has already been merged but [2.4.0](https://github.com/bitcoindevkit/bdk_wallet/issues/426) hasn't been released yet.

The [Wallet 2.4.0](https://crates.io/crates/bdk_wallet/2.4.0) lib was just published and should fix this issue

src/handlers.rs
tvpeter: While this will make it cleaner, I think it is not worth the cost of pulling in a whole library. I'm always for using a minimal number of external libraries.

vadim-anfv: We already pull in `tap` as a dependency in #225: https://github.com/bitcoindevkit/bdk-cli/blob/0a771ca9961da939b13743bfa1b52445db8c20ea/Cargo.toml#L30
So this change doesn't add any extra cost. 

That said, if you'd prefer, we can drop the dependency.

vadim-anfv (approved):
ACK 5def41f63444253556ddab7afe63feffdb62e2c2

Also noted that the PR description mentions "Update bdk_electrum to v0.24.0", but I don't see it in the changes - Cargo.toml (and Cargo.lock) still have bdk_electrum 0.23.2.

tvpeter:
> ACK [5def41f](https://github.com/bitcoindevkit/bdk-cli/commit/5def41f63444253556ddab7afe63feffdb62e2c2)
> 
> Also noted that the PR description mentions "Update bdk_electrum to v0.24.0", but I don't see it in the changes - Cargo.toml (and Cargo.lock) still have bdk_electrum 0.23.2.

Seen the reason now.  I amended the last commit and did not push. Thanks for catching that.

src/handlers.rs
tvpeter: Maybe we can drop in another PR. Generally prefer to avoid deps cost.

PR #256, opened 2026-03-18
fix(test): Enable using any client in config test

### LINKED ISSUE #255: Fix compiling Esplora client out of config test
While building `bdk-cli` inside a flake, I've got the following error:
```bash
> cargoCheckHook flags: -j 16 --profile release --features=repl\,sqlite\,rpc\,sp --target x86_64-unknown-linux-gnu --offline --
>    Compiling bdk-cli v2.0.0 (/build/gbaaz3igwf8dv2785f5vzxhnmhh5g41g-source)
> error[E0599]: no variant or associated item named `Esplora` found for enum `commands::ClientType` in the current scope
>    --> src/config.rs:211:50
>     |
> 211 |         assert_eq!(opts.client_type, ClientType::Esplora);
>     |                                                  ^^^^^^^ variant or associated item not found in `commands::ClientType`
>     |
>    ::: src/commands.rs:188:1
>     |
> 188 | pub enum ClientType {
>     | ------------------- variant or associated item `Esplora` not found for this enum
>
```

I'm not fully aware of what was being tested here, but this is causing cross incompatibilities with the features enabled for the enum and the features enabled for the assert.

_Originally posted by @nymius in https://github.com/bitcoindevkit/bdk-cli/pull/203#discussion_r2950019313_

tvpeter:
### Description

As pointed out in issue #255, whenever Esplora was configured out, the config test will fail because Esplora was used as the client.

This fix addresses the above issue by adding other client types in the test.

Fixes #255 

#### All Submissions:

#### Bugfixes:

* [x] I've added tests to reproduce the issue which are now passing

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -166,20 +166,49 @@ impl TryFrom<&WalletConfigInner> for WalletOpts {
 mod tests {
     use super::*;
     use std::convert::TryInto;
+    const EXT_DESCRIPTOR: &str = "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/0/*)#429nsxmg";
+    const INT_DESCRIPTOR: &str = "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/1/*)#y7qjdnts";
 
     #[test]
     fn test_wallet_config_inner_to_opts_conversion() {
+        #[cfg(any(
+            feature = "electrum",
+            feature = "esplora",
+            feature = "rpc",
+            feature = "cbf"
+        ))]
+        let client_type = {
+            if cfg!(feature = "esplora") {
+                Some("esplora".to_string())
+            } else if cfg!(feature = "rpc") {
+                Some("rpc".to_string())
+            } else if cfg!(feature = "electrum") {
+                Some("electrum".to_string())
+            } else if cfg!(feature = "cbf") {
+                Some("cbf".to_string())
+            } else {
+                None
+            }
+        };
+
         let wallet_config = WalletConfigInner {
             wallet: "test_wallet".to_string(),
             network: "testnet4".to_string(),
-            ext_descriptor: "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/0/*)#429nsxmg".to_string(),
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -169,21 +169,43 @@ mod tests {
...
             #[cfg(any(feature = "electrum", feature = "esplora", feature = "rpc", feature = "cbf"))]
-            client_type: Some("esplora".to_string()),
+            client_type,
+
             #[cfg(any(feature = "electrum", feature = "esplora", feature = "rpc"))]
-            server_url: Some(" https://blockstream.info/testnet4/api".to_string()),
+            server_url: Some("https://example.com/testnet/api".to_string()),
             #[cfg(feature = "electrum")]
-            batch_size: None,
+            batch_size: Some(10),
```
nymius: This is unnecessary as when compiling with electrum `Some(10)` is the default value that replaces the `None` value when calling `WalletOpts::try_into`.

src/config.rs
```diff
@@ -169,21 +169,43 @@ mod tests {
 
     #[test]
     fn test_wallet_config_inner_to_opts_conversion() {
+        #[cfg(any(
```
nymius, line 174: Is this necessary if using `cfg!` macro?

src/config.rs
```diff
@@ -169,21 +169,43 @@ mod tests {
...
+
             #[cfg(any(feature = "electrum", feature = "esplora", feature = "rpc"))]
-            server_url: Some(" https://blockstream.info/testnet4/api".to_string()),
+            server_url: Some("https://example.com/testnet/api".to_string()),
             #[cfg(feature = "electrum")]
-            batch_size: None,
+            batch_size: Some(10),
             #[cfg(feature = "esplora")]
-            parallel_requests: None,
+            parallel_requests: Some(5),
```
nymius: Same than above, `Some(5)` is the default value that replaces `None` value when calling `WalletOpts::try_into` if feature `esplora` is activated.

src/config.rs
```diff
@@ -197,22 +219,18 @@ mod tests {
...
-        ))]
+        #[cfg(feature = "esplora")]
         assert_eq!(opts.client_type, ClientType::Esplora);
 
         #[cfg(feature = "sqlite")]
         assert_eq!(opts.database_type, DatabaseType::Sqlite);
 
+        assert_eq!(
+            opts.ext_descriptor,
+            "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/0/*)#429nsxmg"
```
nymius: This would be a good moment to place these descriptors in `const`s at the beginning of the test module.

src/config.rs
tvpeter: Yes, else the clippy complains and errors out

src/config.rs
```diff
@@ -197,22 +226,15 @@ mod tests {
...
-            "wpkh([07234a14/84'/1'/0']tpubDCSgT6PaVLQH9h2TAxKryhvkEurUBcYRJc9dhTcMDyahhWiMWfEWvQQX89yaw7w7XU8bcVujoALfxq59VkFATri3Cxm5mkp9kfHfRFDckEh/0/*)#429nsxmg"
-        );
 
-        #[cfg(any(
-            feature = "electrum",
-            feature = "esplora",
-            feature = "rpc",
-            feature = "cbf"
-        ))]
+        #[cfg(feature = "esplora")]
```
nymius: Could you reuse something similar than the `if {} else {}` in line 180 to check each of the `client_types` with asserts?

tvpeter: unfortunately `if ...else` blocks did did not work when testing all features.

PR #266, opened 2026-03-31
Feat: Add bip329 wallet labels

### LINKED ISSUE #184: Add BIP-329 wallet label support
**Describe the enhancement**  

This enhancement aims to integrate BIP-329 ("Wallet Labels Export Format") support into `bdk-cli`. This will allow users to:

1.  **Set and manage** human-readable labels for wallet items such as transactions (TXIDs), addresses, and UTXOs (outputs).
2.  **Import labels** from other BIP-329 compliant wallets or previously exported `bdk-cli` label files.
3.  **Export labels** in the standard BIP-329 JSONL format for backup or use in other compatible wallet software.
4.  **View these labels** alongside relevant items in existing `bdk-cli` list commands (e.g., `transactions`, `unspent`).

The implementation will leverage the external `bip329` crate for parsing, managing, and persisting label data in a separate `.jsonl` file (e.g., `<wallet_name>.labels.jsonl`) within the wallet's data directory.

**Use case**  
 

Users of `bdk-cli` will benefit from this feature in several ways:
 
*   **Improved Wallet Management:** Instead of relying solely on cryptic identifiers (TXIDs, long addresses), users can add descriptive labels (e.g., "Payment from Steve," "Savings Address," "Coinjoin UTXO"). This significantly improves transaction history comprehension and overall wallet organization.
*   **Enhanced Interoperability:** By adhering to the BIP-329 standard, users can easily migrate their label data to or from other Bitcoin wallets that support this BIP, reducing vendor lock-in and improving the portability of their wallet metadata.
*   **Better Record Keeping:** Labels provide context for financial activities, which can be useful for personal accounting or tax purposes.
*   **Streamlined Debugging/Testing:** Developers using `bdk-cli` to test BDK applications can label specific test transactions or UTXOs for easier identification and tracking.

**Additional context**
 

*   This feature request was originally discussed in [bdk_wallet issue 168](https://github.com/bitcoindevkit/bdk_wallet/issues/168).
*   The current proposed approach by @notmandatory (https://github.com/bitcoindevkit/bdk_wallet/issues/168#issuecomment-2819699276) is to implement this in `bdk-cli`.
*   The implementation will use the existing [bip329 crate](https://crates.io/crates/bip329).

Musab1258:
### Description

This PR adds BIP-329 wallet label support to `bdk-cli`, to fix #184. It uses the [`[bip329]`](https://github.com/bitcoinppl/bip329) crate to manage label data, persists it in a separate `labels.jsonl` file within the wallet's data directory (e.g. `~/.bdk-bitcoin/<wallet_name>/labels.jsonl`), and keeps it decoupled from the wallet's SQLite/redb database. It also exports and imports label files.

This PR includes the following functionality:

- **Set labels**: It uses a new `label` offline subcommand to attach a human-readable label to either a transaction ID, address, or UTXO (outpoint).
- **Persist labels**: It ensures that Labels are loaded from disk at the start of every offline wallet command and saved back after the command completes, so they survive across sessions.
- **Display labels**: The `unspent` and `transactions` commands now include a `Label` column in both `--pretty` table output and standard JSON output, showing the stored label or an em dash if none exists.
- **Import labels**: The `import_labels` subcommand merges labels from an external BIP-329 JSONL file into the wallet's current label state. If a label for the same item already exists, it is overwritten by the imported one.
- **Export labels**: The `export_labels` subcommand writes the wallet's current label state to a user-specified BIP-329 JSONL file, suitable for backup or use in other BIP-329 compliant wallets.

### Notes to the reviewers

- Labels are stored in a separate `labels.jsonl` file rather than inside the wallet database. This keeps the label format portable and interoperable with other BIP-329 compliant wallets.
- The `Label` import from `bip329` is aliased as `BipLabel` in `handlers.rs` to avoid a name clash with the `OfflineWalletSubCommand::Label` variant brought into scope via `use crate::commands::OfflineWalletSubCommand::*`.
- The `import_labels` command uses the existing `add_or_update_label` helper to merge incoming labels one at a time, new label references are inserted, and existing label references are overwritten by the imported value.
- The `export_labels` command writes to a user-specified path, which is independent of the wallet's internal `labels.jsonl` file. The internal file is always managed automatically by the save-on-exit mechanism, so the two files remain separate.

### Added
- New `label` subcommand to `OfflineWalletSubCommand` for tagging transactions, addresses, and UTXOs with human-readable labels.
- New `import_labels` subcommand to merge labels from an external BIP-329 JSONL file into the wallet's current label state.
- New `export_labels` subcommand to back up the wallet's current label state to a user-specified BIP-329 JSONL file.
- BIP-329 compliant label persistence via a `labels.jsonl` file in the wallet data directory.
- `Label` column in `unspent` and `transactions` output (both `--pretty` and JSON).
- `parse_txid` helper in `utils.rs`.
- `bip329 = "0.4.0"` dependency in `Cargo.toml`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -231,6 +231,50 @@ pub fn handle_offline_wallet_subcommand(
                 }))?)
             }
         }
+        ImportLabels { file } => {
+            let imported_labels = bip329::Labels::try_from_file(&file).map_err(|e| {
+                Error::Generic(format!(
+                    "Failed to import labels from {}: {e}",
+                    file.display()
+                ))
+            })?;
+
+            for label in imported_labels.into_iter() {
+                add_or_update_label(labels, label);
+            }
+
+            if cli_opts.pretty {
+                Ok(format!(
+                    "Successfully imported labels from '{}'",
+                    file.display()
+                ))
+            } else {
+                Ok(serde_json::to_string_pretty(&json!({
+                    "message": "Labels successfully imported",
+                    "file": file.display().to_string()
+                }))?)
+            }
+        }
+        ExportLabels { file } => {
+            labels.export_to_file(&file).map_err(|e| {
+                Error::Generic(format!(
+                    "Failed to export labels to {}: {e}",
+                    file.display()
+                ))
+            })?;
+
+            if cli_opts.pretty {
+                Ok(format!(
+                    "Labels successfully exported to '{}'",
+                    file.display()
+                ))
+            } else {
+                Ok(serde_json::to_string_pretty(&json!({
+                    "message": "Labels successfully exported",
+                    "file": file.display().to_string()
+                }))?)
+            }
+        }
         Unspent => {
             let utxos = wallet.list_unspent().collect::<Vec<_>>();
             if cli_opts.pretty {
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -92,12 +92,14 @@ use {
 };
 
 use bip329::LabelRef;
+use bip329::{AddressRecord, Label as BipLabel, OutputRecord, TransactionRecord};
+use std::ops::DerefMut;
 
 /// Helper to format BIP-329 LabelRefs into searchable strings
 fn format_ref_str(item_ref: &LabelRef) -> String {
     match item_ref {
         LabelRef::Txid(txid) => format!("txid:{}", txid),
-        LabelRef::Address(addr) => format!("addr:{}", addr),
+        LabelRef::Address(addr) => format!("addr:{}", addr.assume_checked_ref()),
         LabelRef::Output(op) => format!("output:{}", op),
         LabelRef::Input(op) => format!("input:{}", op),
         _ => item_ref.to_string(), // Fallback for pubkey/xpub
@@ -106,10 +108,21 @@ fn format_ref_str(item_ref: &LabelRef) -> String {
 
 /// Helper to safely extract a label string or return an em dash
 fn get_label_string(labels: &bip329::Labels, target_ref_str: &str) -> String {
-    labels.iter()
+    labels
+        .iter()
         .find(|l| format_ref_str(&l.ref_()) == target_ref_str)
         .and_then(|l| l.label().map(|s| s.to_string()))
         .unwrap_or_else(|| "—".to_string())
+}
+
+fn add_or_update_label(labels: &mut bip329::Labels, new_label: BipLabel) {
+    let target_ref = new_label.ref_();
+    let labels_vec = labels.deref_mut();
+
+    match labels_vec.iter_mut().find(|l| l.ref_() == target_ref) {
+        Some(existing) => *existing = new_label,
+        None => labels_vec.push(new_label),
+    }
 }
 
 #[cfg(feature = "compiler")]
@@ -178,6 +191,46 @@ pub fn handle_offline_wallet_subcommand(
                 }))?)
             }
         }
+        Label {
+            label_str,
+            address,
+            txid,
+            utxo,
+        } => {
+            let new_label = if let Some(addr) = address {
+                BipLabel::Address(AddressRecord {
+                    ref_: addr.into_unchecked(),
+                    label: Some(label_str.clone()),
+                })
+            } else if let Some(tx) = txid {
+                BipLabel::Transaction(TransactionRecord {
+                    ref_: tx,
+                    label: Some(label_str.clone()),
+                    origin: None,
+                })
+            } else if let Some(outpoint) = utxo {
+                BipLabel::Output(OutputRecord {
+                    ref_: outpoint,
+                    label: Some(label_str.clone()),
+                    spendable: false,
+                })
+            } else {
+                return Err(Error::Generic(
+                    "No target provided for the label".to_string(),
+                ));
+            };
+
+            add_or_update_label(labels, new_label);
+
+            if cli_opts.pretty {
+                Ok(format!("Successfully applied label '{}'", label_str))
+            } else {
+                Ok(serde_json::to_string_pretty(&json!({
+                    "message": "Label successfully applied",
+                    "label": label_str
+                }))?)
+            }
+        }
         Unspent => {
             let utxos = wallet.list_unspent().collect::<Vec<_>>();
             if cli_opts.pretty {
@@ -226,17 +279,23 @@ pub fn handle_offline_wallet_subcommand(
                         "Index".cell().bold(true),
                         "Block Height".cell().bold(true),
                         "Block Hash".cell().bold(true),
-                        "Label".cell().bold(true)
+                        "Label".cell().bold(true),
                     ])
                     .display()
                     .map_err(|e| Error::Generic(e.to_string()))?;
                 Ok(format!("{table}"))
             } else {
-                let utxos_json: Vec<_> = utxos.iter().map(|utxo| {
-                    let mut json_obj = serde_json::to_value(utxo).unwrap();
-                    json_obj["label"] = json!(get_label_string(labels, &format!("output:{}", utxo.outpoint)));
-                    json_obj
-                }).collect();
+                let utxos_json: Vec<_> = utxos
+                    .iter()
+                    .map(|utxo| {
+                        let mut json_obj = serde_json::to_value(utxo).unwrap();
+                        json_obj["label"] = json!(get_label_string(
+                            labels,
+                            &format!("output:{}", utxo.outpoint)
+                        ));
+                        json_obj
+                    })
+                    .collect();
                 Ok(serde_json::to_string_pretty(&utxos_json)?)
             }
         }
@@ -1308,13 +1367,16 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
             let datadir = cli_opts.datadir.clone();
             let home_dir = prepare_home_dir(datadir)?;
             let (wallet_opts, network) = load_wallet_config(&home_dir, &wallet_name)?;
-
             let database_path = prepare_wallet_db_dir(&home_dir, &wallet_name)?;
             let label_file_path = database_path.join("labels.jsonl");
 
             let mut labels = match bip329::Labels::try_from_file(&label_file_path) {
                 Ok(loaded_labels) => loaded_labels,
-                Err(bip329::error::ParseError::FileReadError(io_err)) if io_err.kind() == std::io::ErrorKind::NotFound => bip329::Labels::default(),
+                Err(bip329::error::ParseError::FileReadError(io_err))
+                    if io_err.kind() == std::io::ErrorKind::NotFound =>
+                {
+                    bip329::Labels::default()
+                }
                 Err(e) => return Err(Error::Generic(format!("Failed to load labels: {}", e))),
             };
 
@@ -1363,7 +1425,9 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     &mut labels,
                 )?
             };
-            labels.export_to_file(&label_file_path).map_err(|e| Error::Generic(format!("Failed to save labels: {}", e)))?;
+            labels
+                .export_to_file(&label_file_path)
+                .map_err(|e| Error::Generic(format!("Failed to save labels: {}", e)))?;
 
             Ok(result)
         }
@@ -1521,9 +1585,15 @@ async fn respond(
         ReplSubCommand::Wallet {
             subcommand: WalletSubCommand::OfflineWalletSubCommand(offline_subcommand),
         } => {
-            let value =
-                handle_offline_wallet_subcommand(wallet, wallet_opts, cli_opts, offline_subcommand)
-                    .map_err(|e| e.to_string())?;
+            let mut labels = bip329::Labels::default();
+            let value = handle_offline_wallet_subcommand(
+                wallet,
+                wallet_opts,
+                cli_opts,
+                offline_subcommand,
+                &mut labels,
+            )
+            .map_err(|e| e.to_string())?;
             Some(value)
         }
         ReplSubCommand::Wallet {
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -103,6 +103,7 @@ pub fn handle_offline_wallet_subcommand(
     wallet_opts: &WalletOpts,
     cli_opts: &CliOpts,
     offline_subcommand: OfflineWalletSubCommand,
+    labels: &mut bip329::Labels,
 ) -> Result<String, Error> {
     match offline_subcommand {
         NewAddress => {
@@ -1273,6 +1274,15 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
             let home_dir = prepare_home_dir(datadir)?;
             let (wallet_opts, network) = load_wallet_config(&home_dir, &wallet_name)?;
 
+            let database_path = prepare_wallet_db_dir(&home_dir, &wallet_name)?;
+            let label_file_path = database_path.join("labels.jsonl");
+
+            let mut labels = match bip329::Labels::try_from_file(&label_file_path) {
+                Ok(loaded_labels) => loaded_labels,
+                Err(bip329::error::ParseError::FileReadError(io_err)) if io_err.kind() == std::io::ErrorKind::NotFound => bip329::Labels::default(),
+                Err(e) => return Err(Error::Generic(format!("Failed to load labels: {}", e))),
+            };
+
             #[cfg(any(feature = "sqlite", feature = "redb"))]
             let result = {
                 let mut persister: Persister = match &wallet_opts.database_type {
@@ -1302,6 +1312,7 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     &wallet_opts,
                     &cli_opts,
                     offline_subcommand.clone(),
+                    &mut labels,
                 )?;
                 wallet.persist(&mut persister)?;
                 result
@@ -1314,8 +1325,11 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
                     &wallet_opts,
                     &cli_opts,
                     offline_subcommand.clone(),
+                    &mut labels,
                 )?
             };
+            labels.export_to_file(&label_file_path).map_err(|e| Error::Generic(format!("Failed to save labels: {}", e)))?;
+            
             Ok(result)
         }
         CliSubCommand::Wallet {
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -98,10 +98,10 @@ use std::ops::DerefMut;
 /// Helper to format BIP-329 LabelRefs into searchable strings
 fn format_ref_str(item_ref: &LabelRef) -> String {
     match item_ref {
-        LabelRef::Txid(txid) => format!("txid:{}", txid),
+        LabelRef::Txid(txid) => format!("txid:{txid}"),
         LabelRef::Address(addr) => format!("addr:{}", addr.assume_checked_ref()),
-        LabelRef::Output(op) => format!("output:{}", op),
-        LabelRef::Input(op) => format!("input:{}", op),
+        LabelRef::Output(op) => format!("output:{op}"),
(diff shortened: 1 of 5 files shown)


Musab1258:
Hi @tvpeter, I am done with this. You can check it out when you are free.

I will start working on the follow-up PR to add import and export of label files.

Musab1258:
Hi @tvpeter, I eventually decided to add the import and export of BIP-329 JSONL files' implementation here when, after I was done with the implementation, I saw that it was something that could fit into this PR without bloating it.

So, instead of creating a second PR and having reviewers go through the stress of reviewing two PRs that can probably fit into one, I decided to make it one PR.

notmandatory:
This looks like a very useful feature and one that I'd love to see in the `bdk-cli` project. But instead of coding it directly into `bdk-cli` how would you feel about creating a new `bdk_bip329` (or `bdk-bip329`) crate similar to [bdk-bip322](https://github.com/bitcoindevkit/bdk-bip322)? 

It will be a little more complicated than `bdk-bip322` since you also need some sort of persistence trait to tie into the user's choice of database. But it would be nice to have something easy to integrate with any `bdk_wallet` based project.

Musab1258:
Thanks, @notmandatory, for the review and the suggestion! I really like the idea of creating a new `bdk_bip329` crate similar to `bdk-bip322` so other bdk_wallet projects can easily integrate it, and I would love to work on it.

Regarding the persistence trait to tie into the user's choice of database, I was thinking we could build an interface that lets users store labels internally in their database of choice (like SQLite or Redb). To keep it efficient, we could mirror bdk_wallet's approach by abstracting the backend and using a changeset pattern to persist only the diffs.

To maintain strict BIP-329 compliance, the crate would then provide dedicated import and export methods to convert the internal database state to and from the standard JSONL format.

Does this approach align with what you have in mind?

notmandatory:
Yes this sounds like a good approach.

brh28:
Hey @Musab1258, glad to see you're working on this. I've been playing around with bdk-cli and had some thoughts around this

My thinking is to create a `bdk_labels` tool. The motivation is similar to the idea of a `bdk_bip329` crate, but with some expanded functionality. Some ideas:

1. Import bip329 jsonl (e.g `cat mywallet.jsonl | bdk_labels import --bip329`)
2. Import existing bdk wallets (`bdk_labels import --config # reads $DATADIR, else ~/.bdk-bitcoin`)
3. Add a label to existing wallet using bip329 keys as arguments (e.g `bdk-labels add --type tx --ref <txid> --label "groceries"`)
4. Add a label to types not-associated with a wallet. (e.g `bdk-labels add --type addr --ref <addr> --label "Satoshi's Coins"`) 
5. Export bip329 with filter arguments (`bdk_labels export --origin multisigwallet`)
6. Support a list of labels (or "tags" to distinguish nomenclature) (e.g `bdk-labels tag --type output --ref <txid:3> "cost_basis_lot:20260426" "project_income" "kyc"`). These tags can be serialized into the bip329 label
7. Query labels/tags (e.g `bdk_labels list --type tx --tag project_income`)

Just brainstorming somethings that I think would be useful. Would love to hear know your thoughts

Musab1258:
@notmandatory, since I don't have permissions to create a repository in the bitcoindevkit organization, what should the next steps be? Would you set up an empty repository within the organization (similar to how bdk-bip322 is set up) for me to work from?

> Yes this sounds like a good approach.

Musab1258:
@brh28, these are really nice ideas for expanding the labeling functionality. 

Actually, some of the features you mentioned (like importing a BIP-329 JSONL file and adding labels to existing wallet items) are already implemented in my current PR here. Since @notmandatory suggested implementing it in a new bdk_bip329 crate, we can implement the existing add, import, and export functionalities first. Then we can add the multi-tagging and filter logic later.

Also, the examples you made (like cat mywallet.jsonl | bdk_labels import --bip329 and the rest) show that you wanted the tool to be a CLI. But, I think what @notmandatory wants is a crate that can be integrated into other crates and CLI tools

@notmandatory may also have something to say about this.

> Hey @Musab1258, glad to see you're working on this. I've been playing around with bdk-cli and had some thoughts around this
> 
> My thinking is to create a `bdk_labels` tool. The motivation is similar to the idea of a `bdk_bip329` crate, but with some expanded functionality. Some ideas:
> 
> 1. Import bip329 jsonl (e.g `cat mywallet.jsonl | bdk_labels import --bip329`)
> 2. Import existing bdk wallets (`bdk_labels import --config # reads $DATADIR, else ~/.bdk-bitcoin`)
> 3. Add a label to existing wallet using bip329 keys as arguments (e.g `bdk-labels add --type tx --ref <txid> --label "groceries"`)
> 4. Add a label to types not-associated with a wallet. (e.g `bdk-labels add --type addr --ref <addr> --label "Satoshi's Coins"`)
> 5. Export bip329 with filter arguments (`bdk_labels export --origin multisigwallet`)
> 6. Support a list of labels (or "tags" to distinguish nomenclature) (e.g `bdk-labels tag --type output --ref <txid:3> "cost_basis_lot:20260426" "project_income" "kyc"`). These tags can be serialized into the bip329 label
> 7. Query labels/tags (e.g `bdk_labels list --type tx --tag project_income`)
> 
> Just brainstorming somethings that I think would be useful. Would love to hear know your thoughts

brh28:
Thanks @Musab1258 

I created my own repo [bdk-labels](https://github.com/brh28/bdk-labels) to start implementing some of these features. As you pointed out, many of the features overlap with what you're working on, so it's quite similar to your PR, but perhaps it'll be of interest to you.

The biggest feature difference is `tags`, which are currently encoded into the `label` field but would ideally be it's own bip329 field.

~~Also, my implementation uses an existing [bip329](https://github.com/bitcoinppl/bip329) crate if that's something that interests you.~~ you're already using it :)

src/handlers.rs
```diff
@@ -91,6 +91,40 @@ use {
     bdk_wallet::chain::{BlockId, CanonicalizationParams, CheckPoint},
 };
 
+use bip329::LabelRef;
+use bip329::{AddressRecord, Label as BipLabel, OutputRecord, TransactionRecord};
+use std::ops::DerefMut;
+
+/// Helper to format BIP-329 LabelRefs into searchable strings
+fn format_ref_str(item_ref: &LabelRef) -> String {
```
brh28: should this match the [bip329 spec](https://bips.dev/329/)? if so, it's missing a few types

src/handlers.rs
```diff
@@ -91,6 +91,40 @@ use {
...
+    match item_ref {
+        LabelRef::Txid(txid) => format!("txid:{txid}"),
+        LabelRef::Address(addr) => format!("addr:{}", addr.assume_checked_ref()),
+        LabelRef::Output(op) => format!("output:{op}"),
+        LabelRef::Input(op) => format!("input:{op}"),
+        _ => item_ref.to_string(), // Fallback for pubkey/xpub
```
brh28, line 105: why fallback, rather than handle other types?

notmandatory:
> @notmandatory, since I don't have permissions to create a repository in the bitcoindevkit organization, what should the next steps be? Would you set up an empty repository within the organization (similar to how bdk-bip322 is set up) for me to work from?

Go ahead and create a new repo in your personal account and we can move it into the bitcoindevkit org when it's ready. I also prefer @brh28 's  repo name  `bdk-labels`. And glad to see you two collaborating on this project.

notmandatory:
> @brh28, these are really nice ideas for expanding the labeling functionality.
> 
> Actually, some of the features you mentioned (like importing a BIP-329 JSONL file and adding labels to existing wallet items) are already implemented in my current PR here. Since @notmandatory suggested implementing it in a new bdk_bip329 crate, we can implement the existing add, import, and export functionalities first. Then we can add the multi-tagging and filter logic later.

@Musab1258 you're correct that I do have a general purpose library in mind for this feature, not something only for use in `bdk-cli`. But still good to keep in mind how it will be used in this cli project.

src/handlers.rs
```diff
@@ -156,6 +191,90 @@ pub fn handle_offline_wallet_subcommand(
...
+        ImportLabels { file } => {
+            let imported_labels = bip329::Labels::try_from_file(&file).map_err(|e| {
+                Error::Generic(format!(
+                    "Failed to import labels from {}: {e}",
+                    file.display()
+                ))
+            })?;
+
+            for label in imported_labels.into_iter() {
+                add_or_update_label(labels, label);
```
brh28, line 243: To avoid incidental data loss, I would suggest adding an `--override` flag on import. If the flag is not provided, either the command fails, or only new labels are added

Musab1258:
@notmandatory, I have set up the bdk-labels repository on my personal account. Here it is: [bdk-labels](https://github.com/Musab1258/bdk-labels).

@brh28, I would love to add you as a collaborator so we can work together on this crate. I have sent you a collaboration invite. You should receive it in the mail associated with your GitHub account.

When will you be free so we can discuss and plan out the building of the crate?

brh28:
Yeah man, I'll mention here that I'm mainly interested in serving that tagging feature at the moment as it helps with a particular problem I'm having, but I really like the idea of a wallet cli tool, so happy to help where I can

brh28:
@notmandatory Is it just the `bdk-labels` code that you want to exist separately, or are you imagining an entirely different tool? I'm thinking labeling should be part of the bdk-cli wallet. For example, include a label arg on `new_address` command: 
```
bdk-cli wallet new_address --label "from Alice"
```
Perhaps even a new `addresses` command that lists all wallet addresses along with associated labels

notmandatory:
> @notmandatory Is it just the `bdk-labels` code that you want to exist separately, or are you imagining an entirely different tool? I'm thinking labeling should be part of the bdk-cli wallet. For example, include a label arg on `new_address` command:
> 
> ```
> bdk-cli wallet new_address --label "from Alice"
> ```
> 
> Perhaps even a new `addresses` command that lists all wallet addresses along with associated labels

Yes I'm thinking the `bdk-labels` code should be in its own repo. Then we should add commands like you're suggesting to `bdk-cli` using the `bdk-labels` as a dependency that adds the needed functions to the `bdk_wallet::Wallet`.

Musab1258:
Hi @notmandatory, I am done with the implementation, testing, and documentation of the [bdk-labels](https://github.com/Musab1258/bdk-labels). Can you please take a look at it [here](https://github.com/Musab1258/bdk-labels) when you are free?

Musab1258:
Hi @brh28, I am done with the implementation, testing, and documentation of the [bdk-labels](https://github.com/Musab1258/bdk-labels). Can you please take a look at it [here](https://github.com/Musab1258/bdk-labels) when you are free?



> Yeah man, I'll mention here that I'm mainly interested in serving that tagging feature at the moment as it helps with a particular problem I'm having, but I really like the idea of a wallet cli tool, so happy to help where I can

I think the crate is ready for any feature (tagging, for instance) that we might want to add to it.

> Thanks @Musab1258
> 
> I created my own repo [bdk-labels](https://github.com/brh28/bdk-labels) to start implementing some of these features. As you pointed out, many of the features overlap with what you're working on, so it's quite similar to your PR, but perhaps it'll be of interest to you.
> 
> The biggest feature difference is `tags`, which are currently encoded into the `label` field but would ideally be it's own bip329 field.
> 
> ~Also, my implementation uses an existing [bip329](https://github.com/bitcoinppl/bip329) crate if that's something that interests you.~ you're already using it :)


But, since tagging is not BIP 329 compliant, the [bip329](https://crates.io/crates/bip329) crate we are using in this crate does not implement it. I feel like if we are to do it and still make our bdk-labels crate BIP329 compliant, We should build an abstraction layer on top of the normal labeled wallet (in this bdk-label crate), which implements the tagging feature.

My fear is that if we add it directly, the labels in bdk-wallet will not be compatible with other wallets that are BIP329 compliant.

notmandatory:
@thunderbiscuit to keep the discussion in one place I'm commenting on your Musab1258/bdk-labels#1 here.

I like the idea of keeping the labels feature independent of the `Wallet`.  But if we do it this way then we probably don't need a whole `bdk-labels` crate and only need some examples of how to use the [bip329](https://docs.rs/bip329/0.4.0/bip329/index.html) crate on it's own with a `Wallet` (what this PR does).  The `bip329` crate  only saves/loads jsonl files, but maybe that's enough? 

Sticking with the `bdk-labels::LabelledWallet` approach give some advantages of being able to access the inner `Wallet` which is useful for a few things: 

* can validate what you're labeling exists in `Wallet`
* fields like "spendable" can be determined based on the `Wallet` state
* functions can be added that return items (address, tx, utxos) with any corresponding labels

Or we could go all the way and add labels directly as a new field in `Wallet` and load/persist them in the Wallet change sets, using `bip329` just for import/export.

Musab1258:
@thunderbiscuit, 

I made use of the LabelledWallet adapter because I wanted to add a changeset to the wallet created from bdk_wallet (i.e., a wallet with a label changeset). The pattern allows anyone using the label to add a label(s), validate it against the wallet, and store the result in the changeset.

One of the reasons why the changeset is needed in this crate is that it acts as a diff or a staging area. And if it is not present, the crate would have to overwrite the entire database or rewrite the entire JSONL file every single time a user adds a label.

But if I am to remove the adapter pattern, the crate will not be able to attach its LabelChangeset to the user's wallet. Unless I implement the LabelChangeset logic directly in bdk-wallet. Which kind of reduces the essence of this crate.

But I am willing to move forward with any consensus agreement that is made here.

src/handlers.rs
```diff
@@ -91,6 +91,40 @@ use {
...
+    }
+}
+
+/// Helper to safely extract a label string or return an em dash
+fn get_label_string(labels: &bip329::Labels, target_ref_str: &str) -> String {
+    labels
+        .iter()
+        .find(|l| format_ref_str(&l.ref_()) == target_ref_str)
+        .and_then(|l| l.label().map(|s| s.to_string()))
+        .unwrap_or_else(|| "—".to_string())
```
vadim-anfv, line 115: Unlabeled items render as the literal string `—`. User can set a real label of `—` and JSON then can't tell "no label" from a `—` label.

Reproduction (regtest, two received txs, label `—` set on the first only):

```
# 1. both txs render "—", though no labels set yet
$ bdk wallet --wallet label_demo transactions | jq '.[] | {txid, label}'
{ "txid": "5fe2524f0c46bb12767d1865114031cd631578ac05422a767c0edf23f5c0aeaa", "label": "—" }
{ "txid": "55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465", "label": "—" }

# 2. labels file is empty so the "—" above is the default, not a stored label
$ cat labels.jsonl

# 3. set a real "—" label on the first tx only
$ bdk wallet --wallet label_demo label "—" --txid 5fe2524f0c46bb12767d1865114031cd631578ac05422a767c0edf23f5c0aeaa

# 4. file now has exactly one entry, for the first tx
$ cat labels.jsonl
{"type":"tx","ref":"5fe2524f0c46bb12767d1865114031cd631578ac05422a767c0edf23f5c0aeaa","label":"—"}

# 5. output is unchanged - both still "—", labeled and unlabeled indistinguishable
$ bdk wallet --wallet label_demo transactions | jq '.[] | {txid, label}'
{ "txid": "5fe2524f0c46bb12767d1865114031cd631578ac05422a767c0edf23f5c0aeaa", "label": "—" }
{ "txid": "55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465", "label": "—" }
```

Fix: return `Option<String>` (`None` when absent) so JSON serializes `null` for unlabeled and `"—"` only for a real label.

P.S. `bdk` is my custom zsh fn for `bdk-cli` with additional cli-args.

src/handlers.rs
```diff
@@ -156,6 +191,90 @@ pub fn handle_offline_wallet_subcommand(
...
+                BipLabel::Transaction(TransactionRecord {
+                    ref_: tx,
+                    label: Some(label_str.clone()),
+                    origin: None,
+                })
+            } else if let Some(outpoint) = utxo {
+                BipLabel::Output(OutputRecord {
+                    ref_: outpoint,
+                    label: Some(label_str.clone()),
+                    spendable: false,
```
vadim-anfv, line 215: `label --utxo` hardcodes `spendable: false`, which in BIP-329 marks the UTXO as frozen. Naming a coin shouldn't freeze it. And it can't just be flipped to a hardcoded `true` either - any fixed value clobbers the existing `spendable` - a `true` would silently unfreeze a UTXO the user had frozen, on a mere rename. A naming command must not touch `spendable` at all.

bdk-cli ignores the field locally, but BIP-329 is for interchange - importing into Sparrow/Nunchuk/etc. shows the UTXO as frozen.

Reproduced (regtest) - just naming a coin writes `spendable:false`:

```
$ bdk wallet --wallet label_demo label "my coin" --utxo 55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465:0
$ cat labels.jsonl | jq
{ "type": "output", "ref": "55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465:0", "label": "my coin", "spendable": false }
```

Fix: preserve the existing record's `spendable` (read-modify), default to `true` only for a new record - never hardcode it.

src/handlers.rs
```diff
@@ -1472,9 +1629,15 @@ async fn respond(
         ReplSubCommand::Wallet {
             subcommand: WalletSubCommand::OfflineWalletSubCommand(offline_subcommand),
         } => {
-            let value =
-                handle_offline_wallet_subcommand(wallet, wallet_opts, cli_opts, offline_subcommand)
-                    .map_err(|e| e.to_string())?;
+            let mut labels = bip329::Labels::default();
```
vadim-anfv, line 1632: In REPL mode label reading doesn't seem to work - the "my coin" label shows up outside REPL but is empty inside.

```
$ bdk wallet --wallet label_demo unspent | jq '.[] | {outpoint, label}'
{ "outpoint": "55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465:0", "label": "my coin" }

$ bdk repl --wallet label_demo
> wallet unspent
[ { ... "outpoint": "55d2b85dbf49d78c7070d83f20443e694d90a65498d44bbe84f3c7aa0b55a465:0", "label": "—" ... } ]
```

src/handlers.rs
```diff
@@ -1272,6 +1411,18 @@ pub(crate) async fn handle_command(cli_opts: CliOpts) -> Result<String, Error> {
...
+
+            let mut labels = match bip329::Labels::try_from_file(&label_file_path) {
+                Ok(loaded_labels) => loaded_labels,
+                Err(bip329::error::ParseError::FileReadError(io_err))
+                    if io_err.kind() == std::io::ErrorKind::NotFound =>
+                {
+                    bip329::Labels::default()
+                }
+                Err(e) => return Err(Error::Generic(format!("Failed to load labels: {e}"))),
+            };
```
vadim-anfv, line 1425: Noticed that every wallet command loads (and rewrites) labels.jsonl, and the load is a hard error. Combined with `export_to_file` writing the file without a trailing newline, appending a record - the usual way to add a line to JSONL - glues it onto the last line and makes the file invalid. After that any command fails, even ones unrelated to labels like `balance`:

```
$ echo '{"type":"output","ref":"5fe2524f0c46bb12767d1865114031cd631578ac05422a767c0edf23f5c0aeaa:0","label":"x"}' >> labels.jsonl

$ bdk wallet --wallet label_demo balance
[ERROR bdk_cli] Generic error: Failed to load labels: Unable to parse file: trailing characters at line 1 column 129
```

So a labels file touched by any other tool (or a partial write) can block the whole wallet CLI, including commands that have nothing to do with labels. Might be worth a trailing newline on export and/or skipping a bad labels file instead of failing the command.

va-an (changes requested):
@Musab1258 thanks for the work on BIP-329 labels! Left a few inline comments, please take a look.

One more thing: the address case of `label` looks write-only. `label --address` stores an `addr` record in labels.jsonl, but nothing reads it back - `new_address`/`unused_address` don't show a label, and `unspent`/`transactions` only surface utxo and txid labels. So an address label goes into the file and becomes invisible, unlike txid/utxo. Maybe display address labels somewhere, or drop `--address` until they're readable.

Also, don't forget to rebase on master before further changes!

PR #278, opened 2026-04-21
Refactor bdk-cli structure 

tvpeter:
### Description

This PR restructures the bdk-cli codebase to move away from a monolithic handling model towards a modular architecture. The primary goal is to improve code readability, simplify the addition of new features, and decouple core logic into smaller modules.

Fixes #273 

### Notes to the reviewers

While trying to achieve the above-stated goal, I took maximum care to avoid introducing new bugs. As a result, the existing execution logic was preserved as much as possible.

This PR introduces two new subdirectories under `src`:

 - handlers: This contains all the subcommands' execution logic grouped logically according to the top-level commands.  Modules include the `config`, `descriptor`, `key`, `offline`, `online` and `repl` modules. They each contain structures for the subcommand, execution logic, and implementation of the top-level enum. 
 - utils: This subdirectory contains helper functions used across the app. They are also grouped logically according to related functionality. So far, we have the descriptor-related module, the output that formats responses, and the `types` module that defines response structures.
 
 **Other notable changes**

 - main module is retained with added functionality of routing requests to the right module for execution.
 - All client-related functionalities have been moved from various modules into the `client` module, and it defines the enum that holds all the available clients, as well as implements the functionalities such as broadcasting transactions and syncing operations.
 - The `AppCommand` trait that defines how each subcommand should be structured and the `execute` method they should implement.
 - The `AppContext` struct that holds the universal parameters that subcommands may need. Each subcommand's `execute` method accepts this struct and uses the params for its execution.
 - The output module in the util subdirectory attempts to present a uniform presentation layer by having an output trait, the `FormatOutput` trait, which is implemented by all `types` presenting data response to the user.
 - Collapsed the `wallet` subdirectory into the `persister` module that defines all structures and functionalities related to persistence.
 - Introduced a type state pattern for the AppContext: To minimise runtime errors, the `Init`, `OfflineOperations` and `OnlineOperations` state structs were introduced, and AppContext is generic over these states. This moved the validation from runtime to compile time.
- Introduced `FormatOutput::write_out` and made it generic over the Write trait to allow for flexibility in testing
- Extracted all initialisation logic (database, config loading and persistence initialisation) into `runtime` module. Also, added the `RuntimeWallet` enum to return options of a persisted wallet and a standard non-persisted wallet. The WalletRuntime struct is a wrapper that determines the type of wallet needed and initialises it and handles loading of a wallet from config.

**Request-Response Cycle**

The following outlines how requests are handled and responses are generated after introducing changes to the structure:

[...]

PR #291, opened 2026-06-21
fix(create_tx): enforce single recipient for send_all

### LINKED ISSUE #290: create_tx `--send_all` silently ignores all recipients except the first
In handlers.rs, `create_tx`'s `send_all` branch uses only `recipients[0]` and silently drops the rest:

```rust
            if send_all {
                tx_builder.drain_wallet().drain_to(recipients[0].0.clone());
```

No check that exactly one recipient was given. Passing multiple `--to` with `--send_all` drains everything to the first one, no error - although `send_all` is meant for a single recipient (doc: "Requires only one recipient with value 0").

Notably the silent-payments create_tx in the same file already does the right check:

```rust
            if send_all {
                if sp_recipients.len() == 1 && maybe_recipients.is_none() {
                    tx_builder
                        .drain_wallet()
                        .drain_to(sp_recipients[0].get_placeholder_p2tr_spk());
                } 
```

So the regular create_tx is just inconsistent with the SP one.

Fix: in the regular create_tx send_all branch, error out unless exactly one recipient was provided (mirror the SP-create_tx check), instead of indexing recipients[0].

1estart:
Closes #290

### Description

Previously, passing multiple `--to` arguments with `--send_all` would silently drain the wallet to only the first recipient, ignoring the rest. This was inconsistent with the silent-payments `create_tx`, which already validates the recipient count.

Now returns an error unless exactly one recipient is provided.

### Notes to the reviewers

- The fix mirrors the validation logic already present in the silent-payments `create_tx` branch of the same file, so behavior is now consistent between the two code paths.

## Changelog notice
- Fixed create_tx --send_all to reject multiple recipients instead of silently using only the first one

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog info is also documented on the [GitHub releases](https://github.com/bi
 page. See [DEVELOPMENT_CYCLE.md](DEVELOPMENT_CYCLE.md) for more details.
 
 ## [Unreleased]
+ - Fixed `create_tx --send_all` to reject multiple recipients instead of silently using only the first one
 
 ## [3.0.0]
 
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -540,7 +540,13 @@ pub fn handle_offline_wallet_subcommand(
             let mut tx_builder = wallet.build_tx();
 
             if send_all {
-                tx_builder.drain_wallet().drain_to(recipients[0].0.clone());
+                if recipients.len() == 1 {
+                    tx_builder.drain_wallet().drain_to(recipients[0].0.clone());
+                } else {
+                    return Err(Error::Generic(
+                        "Wallet can only be drained to a single output".to_string(),
+                    ));
+                }
             } else {
                 let recipients = recipients
                     .into_iter()


vadim-anfv (changes requested):
The handlers.rs fix looks good. 

As discussed in https://github.com/bitcoindevkit/bdk-cli/pull/225#discussion_r2962729235 we're moving away from this style of testing (driving the binary as a subprocess), so I'd suggest not adding tests in this PR and dropping the new test file. We can add proper coverage as a follow up after #278 and #289 are merged.

1estart:
> The handlers.rs fix looks good.
> 
> As discussed in [#225 (comment)](https://github.com/bitcoindevkit/bdk-cli/pull/225#discussion_r2962729235) we're moving away from this style of testing (driving the binary as a subprocess), so I'd suggest not adding tests in this PR and dropping the new test file. We can add proper coverage as a follow up after #278 and #289 are merged.

Understood. I've kept the changes minimal and dropped the test file as suggested. I can add the tests later once the related PRs are merged.

tvpeter:
ACK 50d248cfc482bf5d7807b84182e7f19b104ab49b

tvpeter (approved):

vadim-anfv (approved):
ACK 50d248cfc482bf5d7807b84182e7f19b104ab49b

PR #294, opened 2026-06-30
feat(utxo-locking)!: Add commands to lock, unlock and list locked UTXOs

### LINKED ISSUE #293: Add Lock/Unlock UTXO Commands
**Describe the enhancement**  

With the introduction of persistent UTXO locking in Wallet API v3.0.0, these (wallet) commands will enable users to lock UTXOs, check the status UTXOs, and unlock UTXOs. These Locked UTXOs are excluded from coin selection and lock status are persisted. 

**Use case**  
 

-  `lock` utxo command will allow users to lock one or more outpoints
-  `unlock` utxo will allow users unlock one or more outpoints
-  `locked` utxo will allow users to list currently locked outpoints
-  updated `unspent` command will include output with an locked status field per UTXO

**Additional context**
 
See Wallet API [v.3.0.0](https://github.com/bitcoindevkit/bdk_wallet/releases/tag/v3.0.0) release notes

tvpeter:
### Description

This PR updates the Wallet API to v3.1.0 and adds wallet subcommands to lock and unlock UTXOs.  Locked outpoints are excluded from coin selection, and the lock state is stored in the wallet. It also updates the `bdk_redb` to v0.2.0 and `bdk_kyoto` to v0.17.0.

Fixes #293 and builds upon #289 and #278
 
## Changelog notice

- Update bdk_wallet to v3.1.0
- Add wallet `lock_utxo` command
- Add wallet `unlock_utxo` command
- Add wallet `locked_uxtos` command
- Update bdk_redb to v0.2.0
- Update bdk_kyoto to v0.17.0
- Replace `Network` enum with `NetworkKind`
- Replace `include_output_redeem_witness_script` with `add_global_xpubs` in TxBuilder 
- Replace `submit_package` with `broadcast_random` for broadcasting transactions in KyotoClient

#### All Submissions:

#### New Features:


Musab1258:
Concept ACK, but I have a few concerns I want to point out.

Musab1258:
I noticed the bip322 feature, its commands, and integration tests have been commented out.

I assume this was because of the hard dependency conflict between bdk_bip322 (which requires bdk_wallet ^2.1.0) and the bdk_wallet v3.1.0 dependency upgrade and that you left the code here to make it easier to restore later. However, leaving large blocks of dead code may lead to code rot.

Instead of commenting it out, could we cleanly remove the bip322 code in an isolated commit and open a tracking issue to reintroduce it? When bdk_bip322 is updated, we can simply git revert that removal commit to restore the work without needing to re-implement the feature from scratch.

tvpeter:
> I noticed the bip322 feature, its commands, and integration tests have been commented out.
> 
> I assume this was because of the hard dependency conflict between bdk_bip322 (which requires bdk_wallet ^2.1.0) and the bdk_wallet v3.1.0 dependency upgrade and that you left the code here to make it easier to restore later. However, leaving large blocks of dead code may lead to code rot.
> 
> Instead of commenting it out, could we cleanly remove the bip322 code in an isolated commit and open a tracking issue to reintroduce it? When bdk_bip322 is updated, we can simply git revert that removal commit to restore the work without needing to re-implement the feature from scratch.

Thank you for your suggestion. The reason the bip322 feature is currently commented out is that the maintainer of bdk_bip322 @aagbotemi, promised to update the library soon when I reached out to him. I commented out the feature to ensure that the CI checks are passing and to get reviews.

src/handlers/offline.rs
```diff
@@ -851,3 +861,66 @@ impl AppCommand<AppContext<OfflineOperations<'_>>> for VerifyMessageCommand {
...
+impl AppCommand<AppContext<OfflineOperations<'_>>> for LockUtxoCommand {
+    type Output = ListResult<String>;
+    fn execute(&self, ctx: &mut AppContext<OfflineOperations<'_>>) -> Result<Self::Output, Error> {
+        let wallet = &mut ctx.state.wallet;
+        for out_point in &self.utxos {
+            if wallet.get_utxo(*out_point).is_none() {
+                eprintln!("warning: {out_point} is not a known wallet UTXO; locking anyway");
+            }
+            wallet.lock_outpoint(*out_point);
+        }
```
Musab1258, line 881: This prints a warning but proceeds to lock the outpoint even when `wallet.get_utxo` returns None. Is this intentional? 

If it is intentional, does locking this unknown outpoint pollute the persisted lock-state DB with a nonexistent record? Since this outpoint will never appear in list_unspent and thus can never be spent by the wallet, will this orphaned lock record sit in the database forever and cause permanent bloat?

Musab1258:
Regarding the tests, I've been considering a few edge cases:

1. What is the expected behavior when you try to unlock an already unlocked UTXO? Can we add a test to ensure the CLI handles this gracefully?

2. The current `test_locked_utxo_excluded_from_coin_selection()` only tests the scenario where the sole UTXO is locked. Could we add a test for when we have multiple UTXOs with only one of them locked? It would enable us to verify that the coin selection algorithm actively selects the unlocked UTXOs while automatically ignoring the locked ones.

src/handlers/offline.rs
tvpeter: What do you mean by "nonexistent record"?  Locking a UTXO does not exclude it from `unspent` list. It displays with a lock is true. Locking simply makes it unavailable for coin selection. 
Screenshot 2026-07-14 at 3 13 06 AM

tvpeter:
> Regarding the tests, I've been considering a few edge cases:
> 
> 1. What is the expected behavior when you try to unlock an already unlocked UTXO? Can we add a test to ensure the CLI handles this gracefully?
> 2. The current `test_locked_utxo_excluded_from_coin_selection()` only tests the scenario where the sole UTXO is locked. Could we add a test for when we have multiple UTXOs with only one of them locked? It would enable us to verify that the coin selection algorithm actively selects the unlocked UTXOs while automatically ignoring the locked ones.

For exhaustive tests that cover all edge cases, check out the UTXO Locking feature in `bdk_wallet`. This is just to demo how users can utilize the feature.

src/handlers/offline.rs
```diff
@@ -851,3 +860,65 @@ impl AppCommand<AppContext<OfflineOperations<'_>>> for VerifyMessageCommand {
         })
     }
 }
+
+#[derive(Parser, Debug, Clone, PartialEq)]
+pub struct LockUtxoCommand {
+    /// Outpoint(s) to lock, format TXID:VOUT.
+    #[arg(env = "TXID:VOUT", long = "utxo", required = true, value_parser = parse_outpoint)]
```
vadim-anfv: `env = "TXID:VOUT"` [tells clap](https://rust.code-maven.com/clap/clap-and-environment-variables) to read this argument from an env var named `TXID:VOUT` when it isn't passed on the command line.

The catch is that an env var with that name can't actually be set: a name containing `:` isn't a valid shell identifier, so `export` rejects it:

```
$ export TXID:VOUT=foo
export: not valid in this context: TXID:VOUT
```

Might be worth renaming the env var to a more shell-compatible name (without `:`).

src/handlers/offline.rs
```diff
@@ -269,7 +278,7 @@ impl AppCommand<AppContext<OfflineOperations<'_>>> for CreateTxCommand {
         }
 
         if self.offline_signer {
-            tx_builder.include_output_redeem_witness_script();
+            tx_builder.add_global_xpubs();
```
vadim-anfv, line 281: This makes tx building (`finish()`) error with [`CreateTxError::MissingKeyOrigin`](https://docs.rs/bdk_wallet/3.1.0/bdk_wallet/error/enum.CreateTxError.html#variant.MissingKeyOrigin) if any descriptor key has no origin - a bare xpub with no `[fingerprint/path]` prefix (a master key is fine).

The previous `include_output_redeem_witness_script()` didn't have this requirement - in the bdk_wallet version currently on master (2.4.0) that flag was actually never read (checked the source), so it was effectively a no-op. `add_global_xpubs()` does populate the xpubs but is stricter, so in some cases it now surfaces this error.

bdk-cli's own descriptors always carry an origin, so they're fine. Just flagging that an imported watch-only descriptor with a bare origin-less xpub (`wpkh(tpub.../0/*)`) would error here where it previously produced a PSBT. Might be worth a docs note about the origin requirement.

(Same swap in `create_sp_tx`, `bump_fee`, `create_dns_tx`.)

src/handlers/offline.rs
vadim-anfv: > If it is intentional, does locking this unknown outpoint pollute the persisted lock-state DB with a nonexistent record?

My understanding - the use case for locking a currently-unknown outpoint is that the user already knows the UTXO but the wallet doesn't yet (it hasn't synced). After a sync the wallet learns the UTXO, so it shows up among the locked ones and the lock takes effect for coin selection.

Since bdk-cli can't know in advance whether a locked outpoint will become a known wallet UTXO later or not, I can't think of a good solution against the possible DB bloat here - filtering out "unknown" outpoints would also drop these legitimate pre-sync locks.

That said, bdk-cli targets technically advanced users, so I'd lean on the assumption that the user knows what they're doing and won't lock arbitrary UTXOs.

vadim-anfv (approved):
utACK 4a889a7891048650b6f7431f1275a0d0e1ea5f76

Left a few comments, non-blocking in my view.

Looks like a typo in the PR description changelog: "Replace `submit_package` with `broadcast_random`" - the code does the reverse (`broadcast_random` -> `submit_package`).

src/handlers/offline.rs
tvpeter: Thank you. I have updated the documentation for the `--offline-signer` flag to reflects that it adds the PSBT_GLOBAL_XPUB to the PSBT.

Musab1258:
tACK 90270f74bf3f3b7b9c96a4e85e911878d6ec3698

I pulled the branch locally, ran the full integration test suite, and manually verified the complete lock/unlock lifecycle against a funded testnet wallet.

I can confirm the following behaviors work exactly as expected:

- Full test suite passes (45/45) with --all-features.
- The CLI correctly warns the user when locking an unknown UTXO.
- The local SQLite backend successfully persists the lock state across different CLI runs.
- Locked UTXOs properly display "is_locked": true in the unspent JSON output.
- Coin selection actively excludes locked UTXOs (correctly throwing an insufficient funds error when the only available UTXO is locked).
- Unlocking the UTXO cleanly removes the lock state and allows create_tx to succeed.

Note: I couldn't find any bip322-specific tests in the test suite (cargo test --all-features bip322 matches 0). Just stating in case we want to look into it and create another issue for that.

src/handlers/offline.rs
Musab1258: > > If it is intentional, does locking this unknown outpoint pollute the persisted lock-state DB with a nonexistent record?
> 
> My understanding - the use case for locking a currently-unknown outpoint is that the user already knows the UTXO but the wallet doesn't yet (it hasn't synced). After a sync the wallet learns the UTXO, so it shows up among the locked ones and the lock takes effect for coin selection.
> 
> Since bdk-cli can't know in advance whether a locked outpoint will become a known wallet UTXO later or not, I can't think of a good solution against the possible DB bloat here - filtering out "unknown" outpoints would also drop these legitimate pre-sync locks.
> 
> That said, bdk-cli targets technically advanced users, so I'd lean on the assumption that the user knows what they're doing and won't lock arbitrary UTXOs.

When I tested locking an unknown UTXO, it actually showed up in the locked UTXO's persisted database. Since it is not attached to the wallet, it is not listed in the wallet's unspent. And when I unlocked the utxo it is removed from the locked UTXO's persisted database. So, like you said, we can always assume that users will always use a UTXO that actually belongs to the wallet. I guess this is non-blocking.

PR#294(manual-test-1)
PR#294(manual-test)

tvpeter: While I do not think locking a UTXO a wallet does not know about bloats the db (because you can unlock it similarly), I also agree that there might not be a good enough reason to lock a UTXO the loaded wallet does not know about. So I will limit it to locking only UTXOs the wallet knows about during the locking operation.

PR #312, opened 2026-08-06
Add command to delete saved wallet config

### LINKED ISSUE #310: Add command to delete saved wallet config
**Describe the enhancement**  

The wallet config command stores the configuration information for a wallet in the `config.toml` file which is then reused on all subsequent wallet operations. However, there is no way to remove/delete a wallet configuration even when the wallet may no longer be needed.

Adding a command to delete a saved wallet configuration will help reduce the number of redundant wallet configurations and improve the amount of data saved in the config file.

**Use case**  
 
- This will allow users to clean up the wallet config file and remove redundant wallet configs.

yan-pi:
### Description

Adds `wallets delete <wallet_name>` to remove a saved wallet configuration.
Saved configurations can be listed with `wallets list`. 

Resolves #310.

### Notes to the reviewers

List and delete use separate handlers and outputs.

Delete only removes the entry from `config.toml`. 
Wallet database files arenot removed.

## Changelog notice

- Added `wallets delete` to remove a saved wallet configuration

#### All Submissions:

#### New Features:


src/handlers/config.rs
```diff
@@ -158,16 +158,27 @@ impl AppCommand<AppContext<Init>> for SaveConfigCommand {
 }
 
 #[derive(Args, Debug, Clone, PartialEq)]
-pub struct ListWalletsCommand;
+pub struct WalletsCommand {
+    /// Delete the saved configuration for the given wallet instead of listing.
```
tvpeter:
```suggestion
    /// Delete the saved configuration for the given wallet.
```

src/commands.rs
```diff
@@ -141,8 +141,8 @@ pub enum CliSubCommand {
     /// This feature is intended for development and testing purposes only.
     Descriptor(DescriptorCommand),
 
-    /// List all saved wallet configurations.
-    Wallets(ListWalletsCommand),
+    /// List all saved wallet configurations, or delete one with `--delete`.
```
tvpeter:
```suggestion
    /// Saved wallet configuration operations.
```

tvpeter:
> Only the `config.toml` entry is affected, wallet data on disk is not touched (dont know if its the correct path).

So `wallet config` only writes `config.toml`. The DB file is created lazily during an actual wallet operation. So deleting a `config` that was never used leaves nothing on disk. However, if the wallet was used, the DB file remains.

tvpeter (changes requested):
I agree with the approach of the `delete` operation but the returned type of `WalletList` is unnecesary. If returning a message is not possible, then it is better to split the `wallets` into something like `wallets list` and `wallets --delete <name>`. So WalletsSubcommand will be the top-level.

src/handlers/config.rs
```diff
@@ -158,16 +158,27 @@ impl AppCommand<AppContext<Init>> for SaveConfigCommand {
...
+        let mut config = WalletConfig::load(&ctx.datadir)?
+            .ok_or_else(|| Error::Generic("No wallets configured yet.".into()))?;
+
+        if let Some(wallet_name) = &self.delete {
+            if config.remove_wallet(wallet_name).is_none() {
+                return Err(Error::Generic(format!(
+                    "Wallet '{wallet_name}' not found in config"
+                )));
+            }
+            config.save(&ctx.datadir)?;
```
vadim-anfv: Non-blocking: deleting the last wallet leaves `config.toml` with an empty `wallets` table, a state that was unreachable before `--delete`. `WalletConfig::load` returns `None` only when the file is missing, so `wallets` then prints `{}` and exits 0, while the same situation on a fresh datadir fails with "No wallets configured yet." and a non-zero exit (`test_list_wallets_empty`).

Probably worth folding into the `wallets list` / `wallets delete` split you are doing anyway.

src/config.rs
```diff
@@ -346,4 +351,61 @@ mod tests {
...
+            proxy: None,
+            #[cfg(any(feature = "electrum", feature = "esplora"))]
+            proxy_auth: None,
+            #[cfg(any(feature = "electrum", feature = "esplora"))]
+            proxy_retries: None,
+            #[cfg(any(feature = "electrum", feature = "esplora"))]
+            proxy_timeout: None,
+            #[cfg(feature = "cbf")]
+            conn_count: None,
+        }
```
vadim-anfv: Nit: `test_wallet_config_inner_to_opts_conversion` and `test_invalid_client_type_fails` still build `WalletConfigInner` by hand, repeating the same ~20 cfg-gated fields as the new helper.

Since you are adding a helper anyway, consider turning it into a small builder and reusing it there too - those two only differ in a few values. Fine to leave for a follow-up.

vadim-anfv (commented):
cACK e9878775bfa6ae1530b9ac874e9263fea008cd4a

Thanks for picking this up. I left a couple of non-blocking comments inline, both of which fold naturally into the `wallets list` / `wallets delete` restructure @tvpeter asked for.

yan-pi:
@tvpeter, that makes sense. Splitting the operations would also fix the return type issue.

Just to confirm, should this become `wallets list` and `wallets delete <name>`, as @vadim-anfv suggested?

tvpeter:
> @tvpeter, that makes sense. Splitting the operations would also fix the return type issue.
> 
> Just to confirm, should this become `wallets list` and `wallets delete <name>`, as @vadim-anfv suggested?

Yes. We both said the same thing.

src/config.rs
yan-pi: The helper is gone in the updated version. The delete path is covered by the
integration tests instead.

src/handlers/config.rs
yan-pi: Handled on latest push.
Deleting the last saved wallet now removes `config.toml`, so
`wallets list` should behave the same as on a fresh datadir.

yan-pi:
Thanks for the first review! 
I forgot to request another round of review. Now might be ready for it.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ page. See [DEVELOPMENT_CYCLE.md](DEVELOPMENT_CYCLE.md) for more details.
 
 ## [Unreleased]
 
+- Split saved wallet configuration operations into `wallets list` and `wallets delete <wallet_name>`
```
vadim-anfv, line 8: This entry and the "Changelog notice" in the PR description are out of sync - could you align them?

While you're here: plain `bdk-cli wallets` no longer works, and imo two entries would sit closer to https://keepachangelog.com/en/1.0.0/ and make that clearer. Up to you though:

```suggestion
- Replaced plain `wallets` with `wallets list`
- Added `wallets delete <wallet_name>` to remove a saved wallet configuration
```

src/commands.rs
```diff
@@ -208,6 +212,16 @@ pub enum CliSubCommand {
     ResolveDnsRecipient(ResolveDnsRecipientCommand),
 }
 
+/// Saved wallet configuration subcommands.
+#[derive(Debug, Subcommand, Clone, PartialEq)]
+pub enum WalletsSubCommand {
+    /// List saved wallet configurations.
+    List(ListWalletsCommand),
+
+    /// Delete a saved wallet configuration.
```
vadim-anfv, line 221: It would be great to have what you wrote in the PR description in `wallets delete --help` as well:

```suggestion
    /// Delete a saved wallet configuration.
    ///
    /// Only the entry in `config.toml` is removed. Wallet data on disk is left in place.
```

vadim-anfv (approved):
tACK f2a7e8f8f4d0cfda7b916a83d3947f12825cfc5b

Both of my earlier comments are addressed. I can't resolve the threads myself - could you close them? 

Added two non-blocking notes inline.

tvpeter (changes requested):
Thank you @yan-pi for working on this issue.

So having reconsidered the approach I suggested earlier, I think it may not be a good idea to delete saved configurations that have wallet data already persisted. So it will be better to handle it in such a way that, if the wallet data exists, return an error that wallet data exist for that configuration and do not delete the config. While it is possible to recreate the same config, the user might have lost the descriptors to do so. Also, if at all further down the line, we decide to consider deleting, it should also wipe the db file (for sqlite and much harder to even consider for redb) with a `--force` flag.

Aside that, kindly address the minor failing checks and review earlier by Vadim.

Thank you

yan-pi:
The failing checks are unrelated to this PR. 
The `all-features` build and Clippy stop during compilation on an unused import in `src/handlers/payjoin/db.rs:537`:

```rust
   use payjoin::persist::SessionPersister as _;
```

This file is not modified by this PR, and no Payjoin tests are executed. The default/no-default feature checks and formatting pass.

PR #315, opened 2026-08-18
chore: remove unused tap dependency

### LINKED ISSUE #314: Use or delete the `tap` dependency
`tap` is declared as a dependency in `Cargo.toml`, but is never used:

```console
$ cargo +nightly udeps --all-features --all-targets
...
unused dependencies:
`bdk-cli v4.0.0 (/home/v/src/bitcoindevkit/bdk-cli)`
└─── dependencies
     └─── "tap"
...
```

Rather than dropping it, I'd suggest putting it to use. Two places where it would make the code read better:

`CompileCommand::execute` in `src/handlers/descriptor.rs`:

```rust
// now
let legacy_policy: Miniscript<String, Legacy> = policy
    .compile()
    .map_err(|e| Error::Generic(e.to_string()))?;
let segwit_policy: Miniscript<String, Segwitv0> = policy
    .compile()
    .map_err(|e| Error::Generic(e.to_string()))?;
let taproot_policy: Miniscript<String, Tap> = policy
    .compile()
    .map_err(|e| Error::Generic(e.to_string()))?;

let descriptor = match self.script_type.as_str() {
    "sh" => Descriptor::new_sh(legacy_policy),
    "wsh" => Descriptor::new_wsh(segwit_policy),
    "sh-wsh" => Descriptor::new_sh_wsh(segwit_policy),
    // ...
}?;

// with tap::Pipe
let descriptor = match self.script_type.as_str() {
    "sh" => policy.compile::<Legacy>()?.pipe(Descriptor::new_sh),
    "wsh" => policy.compile::<Segwitv0>()?.pipe(Descriptor::new_wsh),
    "sh-wsh" => policy.compile::<Segwitv0>()?.pipe(Descriptor::new_sh_wsh),
    // ...
}?;
```

There is also a bug in that spot: the policy is compiled for all three contexts up front, so a policy that only fits the requested type can still be rejected by one of the other two. I don't want to expand this issue with the details, but the rewrite above fixes it as a side effect. If we decide to remove `tap` instead, I'll open a separate issue for the bug.

`BumpFeeCommand::execute` in `src/handlers/offline.rs`:

```rust
// now
let mut tx_builder = wallet.build_fee_bump(self.txid)?;
tx_builder.fee_rate(fee_rate);

if let Some(address) = &self.shrink_address {
    tx_builder.drain_to(address.script_pubkey());
}
if self.offline_signer {
    tx_builder.add_global_xpubs();
}

let psbt = tx_builder.finish()?;

// with tap::Tap
let psbt = wallet
    .build_fee_bump(self.txid)?
    .tap_mut(|b| { b.fee_rate(fee_rate); })
    .tap_mut(|b| if let Some(address) = &self.shrink_address {
        b.drain_to(address.script_pubkey());
    })
    .tap_mut(|b| if self.offline_signer { b.add_global_xpubs(); })
    .finish()?;
```

If that doesn't seem worth a dependency, then the other way is to remove the `tap` from dependencies.
@tvpeter your call - happy to open a PR either way.

vadim-anfv:
Closes #314

### Description

Removes the unused `tap` dependency, per @tvpeter's [decision](https://github.com/bitcoindevkit/bdk-cli/issues/314#issuecomment-5311773042).

## Changelog notice

- Removed the unused `tap` dependency

#### All Submissions:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,7 +269,6 @@ dependencies = [
  "serde",
  "serde_json",
  "shlex 1.3.0",
- "tap",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
@@ -3344,12 +3343,6 @@ dependencies = [
  "quote",
  "syn",
 ]
-
-[[package]]
-name = "tap"
-version = "1.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
 
 [[package]]
 name = "tar"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ tracing = "0.1.44"
 tracing-subscriber = "0.3.20"
 toml = "1.1.0"
 serde= {version = "1.0", features = ["derive"]}
-tap = "1.0.1"
 
 # Optional dependencies
 bdk_bitcoind_rpc = { version = "0.22.0", features = ["std"], optional = true }


tvpeter (approved):
ACK 7e8b960b23f5ac8ec6d300c06adb9345c1615b25

PR #319, opened 2026-09-05
Refractor WalletRuntime load to Combine wallet DB dir setup and config loading into one function

emma31-dev:
### Description

Merges `prepare_wallet_db_dir` and `load_wallet_config` into a single `prepare_wallet_db_dir_and_config` function that returns the database path alongside WalletOpts and Network.

The two functions were always called together in WalletRuntime::load, so combining them removes the duplication and simplifies the call site.

### Notes to the reviewers

The main reason for this change is because the two functions merged were single-use.

#### All Submissions:

Thank you.

### DIFF
--- a/src/utils/common.rs
+++ b/src/utils/common.rs
@@ -112,22 +112,6 @@ pub(crate) fn prepare_home_dir(home_path: Option<PathBuf>) -> Result<PathBuf, Er
     Ok(dir)
 }
 
-/// Prepare wallet database directory.
-#[allow(dead_code)]
-pub(crate) fn prepare_wallet_db_dir(
-    home_path: &Path,
-    wallet_name: &str,
-) -> Result<std::path::PathBuf, Error> {
-    let mut dir = home_path.to_owned();
-    dir.push(wallet_name);
-
-    if !dir.exists() {
-        std::fs::create_dir(&dir).map_err(|e| Error::Generic(e.to_string()))?;
-    }
-
-    Ok(dir)
-}
-
 pub fn is_mnemonic(s: &str) -> bool {
     let word_count = s.split_whitespace().count();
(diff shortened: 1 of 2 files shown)


vadim-anfv (commented):
Thanks for the patch. I don't see the duplication this is meant to remove: both functions were called exactly once, on two consecutive lines of `WalletRuntime::load`. Being single-use isn't in itself a reason to merge two small helpers.

The call site reads worse to me now: two named calls became one returning `(PathBuf, WalletOpts, Network)` unpacked by position. I'd also rather each helper stayed responsible for a single thing, and a `do_x_and_y` name is a smell worth avoiding rather than introducing. Keeping the directory setup behind its own name reads better to me than having its body unfolded into the config loader.

This also isn't a pure refactor, it changes observable behavior. Same command on both, with a wallet name that isn't in the config:

    $ bdk-cli --network testnet --datadir $D wallet --wallet ghost balance
    Error: Generic error: No config found for wallet ghost

    $ ls $D
    master:      (empty)
    this branch: ghost/

The merged function creates the directory before it validates the config, so a failed lookup now leaves an empty directory behind. Moving `create_dir` below the config parsing restores the old behavior, but then the function is just two independent blocks in a row.

emma31-dev:
My bad I didn't pay much attention to ordering. Should I reorder the change and keep it or forget it all together?

vadim-anfv:
> My bad I didn't pay much attention to ordering. Should I reorder the change and keep it or forget it all together?

No worries. My take: I'd drop it. Reordering makes it behavior-neutral again, but then it's two independent blocks in one function and the duplication the description refers to isn't there, so not much is left that the change buys.

I'm not a maintainer though, so treat that as one reviewer's opinion rather than a verdict.

PR #320, opened 2026-09-07
Refractor runtime persister to error out when not persisted

emma31-dev:
### Description

The `persist()` method of `RuntimeWallet` returns `ok(())` wether or not the wallet was persisted or not. This is bad practice. My changes Return an error when wallet persistence fails instead of silently succeeding when no changes were written.

#### All Submissions:

Thank you.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,7 +269,6 @@ dependencies = [
  "serde",
  "serde_json",
  "shlex 1.3.0",
- "tap",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
@@ -3344,12 +3343,6 @@ dependencies = [
  "quote",
  "syn",
 ]
-
-[[package]]
-name = "tap"
-version = "1.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
 
 [[package]]
 name = "tar"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ tracing = "0.1.44"
 tracing-subscriber = "0.3.20"
 toml = "1.1.0"
 serde= {version = "1.0", features = ["derive"]}
-tap = "1.0.1"
 
 # Optional dependencies
 bdk_bitcoind_rpc = { version = "0.22.0", features = ["std"], optional = true }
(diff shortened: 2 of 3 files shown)


vadim-anfv (commented):
`Wallet::persist` returns whether any new changes were written, not whether the write succeeded. A failing persister comes back as `Err` and is already propagated by `?`, so `false` isn't a failure - it means there was nothing staged to write, which is the normal outcome for a command that only reads.

`wallet.persist()` runs after every wallet subcommand in `src/main.rs`, both for the online and the offline branch, so turning `false` into an error makes read-only commands fail after they have already produced correct output:

    $ bdk-cli --network testnet --datadir $D wallet --wallet w balance
    {
      "total": 0,
      "trusted_pending": 0,
      "untrusted_pending": 0,
      "immature": 0,
      "confirmed": 0
    }
    Error: Generic error: Wallet changes were not persisted

    $ echo $?
    1

Same on this branch for `unspent`, `transactions`, `unused_address` and `locked_utxos`. `new_address` passes only because revealing an address stages a change. Before this change the return value was ignored, so none of them could fail this way.

So I don't think this should land as is: the case it converts into an error is the normal one.

emma31-dev:
In the persist function, this happens:
```rust
pub fn persist(&mut self, persister: &mut P) -> Result<bool, P::Error> {
        match self.inner.staged_mut() {
            Some(stage) => {
                P::persist(persister, &*stage)?;
                let _ = stage.take();
                Ok(true)
            }
            None => Ok(false),
        }
    }
```

Where it can still return okay when no change is persisted.

emma31-dev:
Wouldn't a warning when change not staged still be necessary?

vadim-anfv:
> In the persist function, this happens:
> 
> ```rust
> pub fn persist(&mut self, persister: &mut P) -> Result<bool, P::Error> {
>         match self.inner.staged_mut() {
>             Some(stage) => {
>                 P::persist(persister, &*stage)?;
>                 let _ = stage.take();
>                 Ok(true)
>             }
>             None => Ok(false),
>         }
>     }
> ```
> 
> Where it can still return okay when no change is persisted.

Have a look at the [doc comment](https://github.com/bitcoindevkit/bdk_wallet/blob/6fc68462e01aebf346c9f43e04858e83bf209102/src/wallet/persisted.rs#L190) right above that function: "Returns whether any new changes were persisted." The bool only says whether anything was written, so `true` and `false` are both normal outcomes - a real problem comes back as `Err` instead.

> Wouldn't a warning when change not staged still be necessary?

A warning would print for every command that persists nothing, which is the same point as before: for read-only commands that's the expected state, and there's nothing for the user to act on.

emma31-dev:
I understand my previous assumption was made on the wrong working of the persist method. My mistake.

PR #326, opened 2026-09-17
Fix transaction creation input validation and fee rate calculation

### LINKED ISSUE #325: create_tx panics on malformed input instead of returning an error; fee-rate parsing silently truncates or falls back
**Describe the bug**  

`CreateTxCommand` and `BumpFeeCommand` call `.unwrap()` on `Result`s that carry user-supplied input, so a mistyped argument aborts the process with a Rust panic (exit 101) instead of a usage error (exit 1).
Also, related is the fee_rate in `create_tx`, `create_sp_tx`, and `bump_fee` where the `--fee_rate` is taken as `f32` and cast it with
`as u64`. The cast is saturating and truncating, and when `FeeRate::from_sat_per_vb` returns `None` the `if let Some` is simply skipped (or `unwrap_or(FeeRate::BROADCAST_MIN)` in `bump_fee`), so the user silently gets a fee they did not ask for.

**To Reproduce**  

`$ADDR` is any valid testnet address.

```sh
# control — normal error path, exit 1
bdk-cli --network testnet wallet --wallet repro create_tx --to "$ADDR:10000"
# Error: Create transaction error: Insufficient funds: 0 BTC available of 0.00010054 BTC needed

# E1 — unknown outpoint, exit 101
bdk-cli --network testnet wallet --wallet repro create_tx --to "$ADDR:10000" \
  --utxos aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:0
# thread 'main' panicked at src/handlers/offline.rs:291:46:
# called `Result::unwrap()` on an `Err` value: UnknownUtxo(OutPoint { txid: aaaa…, vout: 0 })

# E2 — malformed base64, exit 101
bdk-cli --network testnet wallet --wallet repro create_tx --to "$ADDR:10000" \
  --add_data '!!!not-base64!!!'
# thread 'main' panicked at src/handlers/offline.rs:299:70:
# called `Result::unwrap()` on an `Err` value: InvalidByte(0, 33)
```

**Expected behavior**  

A graceful error on any failure instead of panic

**Build environment**  
 - BDK-CLI tag/commit: v4.0.0 5b3cb00

tvpeter:
### Description

This PR addresses input-validation problems on the transaction-building commands (create_tx, create_sp_tx, bump_fee), transaction fee rate and OP_RETURN data size:
- Panics on malformed input:  `create_tx` and `bump_fee` called `.unwrap()` on Results so they panic (exit 101) instead of an error (exit 1).
-  The `create_sp_tx` already guarded these paths, but improvements were made to the error type been returned
- Fee rates: `--fee_rate` was an `f32` cast with `as u64`, which truncates and saturates, and a `None` from `from_sat_per_vb` was silently skipped. Parsing now happens in a `value_parser`, so bad values are rejected with
a usage message before a wallet is loaded. Because `FeeRate` counts sat/kwu, fractional rates keep 1/250 sat/vB precision instead of being truncated .
- OP_RETURN size: Both `--add_data` and `--add_string` document "max 80 bytes" and neither enforced it. This has now been updated to 100_000 bytes and enforced in transaction building.
- `create_dns_tx` was had the same fee-rate bug and the same OP_RETURN handling, and has been fixed too.
- `bump_fee --utxos` and `send_payjoin -f` are fixed by the same changes. 

Fixes #325 

## Changelog notice

- Fixed `create_tx` and `bump_fee` panicking on malformed `--utxos` and `--add_data` values instead of returning an error
- Fixed `--fee_rate` silently truncating to a whole sat/vB, falling back to a default, or producing a zero-fee transaction; unusable values are now rejected
- Enforced the documented 100_000 byte limit on `--add_data` and `--add_string` OP_RETURN payloads

#### All Submissions:

#### Bugfixes:

* [x] I've added tests to reproduce the issue which are now passing


vadim-anfv (commented):
`MAX_OP_RETURN_BYTES` (99_994) counts the OP_RETURN scriptPubKey only, so `create_tx` builds transactions over the 100_000 vB standard tx size limit, which no default-policy node relays. The output alone is 100_013 vB (8 value + 5 length prefix + 100_000 script), before any input, change or header.

A repro, not a test I'm suggesting you merge:

```rust
/// `--add_string` accepts a payload of up to `MAX_OP_RETURN_BYTES` (99_994),
/// but the resulting transaction is over the 100_000 vB standardness limit,
/// so no default-policy node relays it.
#[test]
fn test_max_op_return_payload_fits_the_standard_tx_size() {
    let (cli, mut cmd_init, env) = setup_online_wallet();
    cmd_init.assert().success();
    fund_and_sync_wallet(&cli, &env);

    let data = "A".repeat(99_994);
    let to = format!("{RECIPIENT}:15000");
    let args = ["create_tx", "--to", &to, "--add_string", &data];
    let psbt = run_wallet_json(&cli, &args)["psbt"].as_str().unwrap().to_owned();

    // Signing it would mean passing ~200 KB of base64 as an argument, and the
    // unsigned tx is enough here: the signed one is only bigger.
    let tx = bdk_wallet::bitcoin::Psbt::from_str(&psbt).unwrap().unsigned_tx;
    let result = env.rpc_client().test_mempool_accept(&[&tx]).unwrap();

    let reason = result[0].reject_reason.as_deref();
    assert!(
        reason != Some("tx-size"),
        "node rejected the {} vB tx built at the documented limit: {}",
        tx.vsize(),
        reason.unwrap_or("accepted")
    );
}
```

```
$ cargo test --all-features --test cli test_max_op_return

test ...::test_max_op_return_payload_fits_the_standard_tx_size ... FAILED

node rejected the 100150 vB tx built at the documented limit: tx-size
```

That tx is 150 vB over the limit with one input, a recipient and change, so the usable payload is at most ~99_844 here, less with more inputs. `the_largest_allowed_payload_fits_the_script_limit` locks in the same unusable size: a 100_000 byte script never fits a standard tx.

Non-blocking: master enforced no limit here at all, so this is an improvement either way.

notmandatory:
> `MAX_OP_RETURN_BYTES` (99_994) counts the OP_RETURN scriptPubKey only, so `create_tx` builds transactions over the 100_000 vB standard tx size limit, which no default-policy node relays. The output alone is 100_013 vB (8 value + 5 length prefix + 100_000 script), before any input, change or header.

Is this something that needs to be checked and enforced in the `bdk_tx` crate?

vadim-anfv:
> Is this something that needs to be checked and enforced in the `bdk_tx` crate?

bdk-cli doesn't use `bdk_tx`, transactions are built with `TxBuilder` from `bdk_wallet`.

Is the plan to move transaction building to `bdk_tx`, in bdk-cli in particular? 
If so, the tx size check belongs there.

notmandatory:
> Is the plan to move transaction building to `bdk_tx`, in bdk-cli in particular? If so, the tx size check belongs there.

The plan is to transition `bdk_wallet` to depreciate `TxBuilder` and use `bdk_tx` via `Wallet::create_psbt` (see https://github.com/bitcoindevkit/bdk_wallet/pull/516). So I'd suggest adding the check to `bdk-tx` and updating `bdk-cli` to enable the `Wallet::create_psbt` experimental feature and using it. But that all will need to be done in new PRs.

vadim-anfv:
> The plan is to transition `bdk_wallet` to depreciate `TxBuilder` and use `bdk_tx` via `Wallet::create_psbt` (see [bitcoindevkit/bdk_wallet#516](https://github.com/bitcoindevkit/bdk_wallet/pull/516)). So I'd suggest adding the check to `bdk-tx` and updating `bdk-cli` to enable the `Wallet::create_psbt` experimental feature and using it. But that all will need to be done in new PRs.

Two things here.

1. Moving to `Wallet::create_psbt`: it is in bdk_wallet [3.2.0](https://github.com/bitcoindevkit/bdk_wallet/releases/tag/v3.2.0), so we can pick it up now. I'll open an issue for it.

2. The size check: 100_000 vB is default relay policy, not consensus, and miners do accept larger txs out of band, so I'd rather bdk_tx didn't refuse to build such a tx. I'd keep the check closer to the user, where refusing or warning is a call the app can make. `create_psbt` is also early for it: an unsigned tx can be under the limit and go over it once the witnesses are there, so the check belongs after signing.

